PR #5, opened 2024-07-28
Minor doc fixes

mik90:
Fixes some typos and adds relative markdown links where possible. It's easier to navigate around docs locally when markdown links are relative.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Copper is a user-friendly robotics framework designed for creating fast and reli
 
 * Reliable: Copper leverages Rust's ownership, type system, and concurrency model to minimize bugs and ensure thread safety.
 
-* Product Oriented: Copper trying hard to avoid late infra related integration issues by generating a very predictable runtime.
+* Product Oriented: Copper aims to avoid late-stage infra integration issues by generating a very predictable runtime.
 
 > [!NOTE]
 > Copper is still in **early development**, and the API is subject to change. We are looking for contributors to help us build the best robotics framework possible. If you are interested, please join us on [Gitter](https://gitter.im/copper-project/copper-rs) or open an issue.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,11 +45,11 @@ It will ask you the name you want to pick interactively.
 cargo install cargo-generate
 git clone https://github.com/copper-project/copper-rs
 cd copper-rs/templates
-cargo nunew [path_where_you_want_your_project_created]
+cargo cunew [path_where_you_want_your_project_created]
     🤷   Project Name:
 ```
 
-Check out [copper-templates](https://github.com/copper-project/copper-rs/templates/README.md) for more info.
+Check out [copper-templates](templates/README.md) for more info.
```
gbin, line 52: Ha I include the readme in the rust crate doc here: https://github.com/copper-project/copper-rs/blob/c0f6d8e1e18891a19a955f7fe7de3c4d5181711b/cu29/src/lib.rs#L1

It means that relative links won't work.

An alternative could be to move the templating into the cu29 crate?

gbin (changes requested):

README.md
mik90: ah gotcha. Would this be a workaround https://gist.github.com/JakeHartnell/2c1fa387f185f5dc46c9429470a2e2be?

I can undo the relative links as well

gbin: oooh yes, this is clever! let's try this trick

mik90: fwiw, without changes, i was able to see the README when running `cargo doc` and opening up the results in firefox

I'll push the workaround since I don't know if it'll stay that way when published to https://copper-project.github.io/copper-rs/cu29/index.html

mik90: Updated in 902afc9dce541477e55c40bc9afa2a87699c6ca1

gbin (approved):

Issue #8, opened 2024-07-29
Testing changes a committed file: cu29_export/test/cu29_log_index/lock.mdb


gbin:
This is one of the very initial tests I made on the log index, now we can easily build an end to end integration test.

PR #13, opened 2024-08-06
export: fix tests changing committed files

gbin:
This is a quick fix for #8 

As this is a test I think it is just ok to work around the change issue.


mik90 (approved):
Workaround looks good

As we talked about in gitter, ideally we'd use an index generated on the fly as opposed to one committed to the repo

Would we be able to remove the lock file from the repo now? Or is that required in order to open the index?

PR #14, opened 2024-08-10
log: Simplified and optimized the structlog

gbin:
Notable change: I completely removed the queue in the runtime as the overhead was larger than a simple unbuffered mutex. 

I also build a quick benchmark to compare the speed with the more traditional Rust text logging.

```
sal ➜  cu_standalone_structlog (gbin/structlog_optim) ✗ source back2back.sh
   Compiling cu_standalone_structlog v0.2.1 (/home/gbin/projects/copper/copper-rs/examples/cu_standalone_structlog)
    Finished `release` profile [optimized] target(s) in 1.92s
     Running `/home/gbin/projects/copper/copper-rs/target/release/standardlog_perf`
Total time: 34.220 s in ./logfile.txt
.rw-r--r-- gbin gbin 101 MB Sat Aug 10 10:45:55 2024  logfile.txt
   Compiling cu_standalone_structlog v0.2.1 (/home/gbin/projects/copper/copper-rs/examples/cu_standalone_structlog)
CLog: ==================================================================================
CLog: Interned strings are stored in: "/home/gbin/projects/copper/copper-rs/target/release/cu29_log_index"
CLog:    [r] is reused index and [n] is new index.
CLog: ==================================================================================
CLog: #001 [r] -> This is the logline {} associated with the log Logging and some more = {}, {}.
    Finished `release` profile [optimized] target(s) in 1.84s
     Running `/home/gbin/projects/copper/copper-rs/target/release/structlog_perf`
Total time: 70.342 ms in ./logfile.bin
.rw-r--r-- gbin gbin 28 MB Sat Aug 10 10:45:57 2024  logfile.bin
```

60% reduction in logged size and 486x faster walltime

### DIFF
--- a/cu29_log/src/lib.rs
+++ b/cu29_log/src/lib.rs
@@ -3,6 +3,7 @@ use cu29_clock::CuTime;
 use cu29_traits::{CuError, CuResult};
 pub use cu29_value as value;
 use serde::{Deserialize, Serialize};
+use smallvec::SmallVec;
 use std::collections::HashMap;
 use std::fmt::Display;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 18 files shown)


cu29_log_runtime/src/lib.rs
```diff
@@ -1,207 +1,248 @@
...
-                read_interned_strings(extra.path_to_index.as_path())
-                    .expect("Failed to read the interned strings"),
-            );
-            let logger = Some(extra.logger.clone());
-            (index, logger)
+        // If WRITER is already initialized, update the inner value.
+        // This should only be useful for unit testing.
+        if let Some((writer, _)) = WRITER.get() {
+            let mut writer_guard = writer.lock().unwrap();
+            *writer_guard = Box::new(destination);
```
mik90, line 44: We could make a separate LoggerRuntime testing function instead of having this part of the default init function. Something like

```rust
#[cfg(test)]
fn overwrite_writer()
```

Really depends on if you want extra init calls to be ignored or the last one to apply. I figure the former is expected behavior

cu29_log_runtime/src/lib.rs
```diff
@@ -1,207 +1,248 @@
...
+}
+
+// #[cfg(test)]
+// mod tests {
+//     use crate::CuLogEntry;
+//     use bincode::config::standard;
+//     use cu29_log::value::Value;
+//
+//     #[test]
+//     fn test_encode_decode_structured_log() {
```
mik90, line 235: Is this test meant to be revived or should it still be disabled?

cu29_log_runtime/src/lib.rs
```diff
@@ -1,207 +1,248 @@
...
+use bincode::config::Configuration;
+use bincode::enc::{Encoder, EncoderImpl};
+use bincode::error::EncodeError;
+use bincode::enc::write::Writer;
+use bincode::enc::Encode;
+#[cfg(debug_assertions)]
+use cu29_intern_strs::read_interned_strings;
 
-// The logging system is basically a global queue.
-static QUEUE: OnceCell<Sender<CuLogEntry>> = OnceCell::new();
```
mik90, line 15: Has this change been run in a benchmark with producers from multiple threads? The two examples are single-threaded producers and I'm wondering if the queue overhead makes it worth it in that scenario.

With a locked logger, callers will block on accessing it but if accessing it is quick, then it may make sense to keep it that way. But if many log producers are contending for the lock, it's possible a mpsc queue would be faster

mik90:
```rust
    for i in 0..1_000_000 {
        info!("This is the logline {} associated with the log Logging and some more = {}, {}", i , i+2, i+3);
    }
```

this does test raw throughput but I'm thinking that it'd be useful to model logging frequency that's not as extreme.

Maybe log every 250us/1ms for a 10ms period, sleep for a bit, then do it again. And create N producers with this configurable behavior. That's certainly more implementation behavior, but I think it'd help with predicting future performance

mik90: I threw together some code to show what i mean

[0001-Multi-threaded-producer.patch](https://github.com/user-attachments/files/16574343/0001-Multi-threaded-producer.patch)

Thoughts?

cu29_log/src/lib.rs
```diff
@@ -15,20 +16,70 @@ const INDEX_DIR_NAME: &str = "cu29_log_index";
 #[allow(dead_code)]
 pub const ANONYMOUS: u32 = 0;
 
+pub const MAX_PARAMS: usize = 10;
```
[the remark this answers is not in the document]
gbin: Done

cu29_log_runtime/src/lib.rs
gbin: oh yes so sorry, I was postponing that because of the smallvecs... fixed!

cu29_log_runtime/src/lib.rs
gbin: I agree we need to test the contention too. Can you make just another pair of executable to test them vs normal text logging?

cu29_log_runtime/src/lib.rs
gbin: Yeah this one is a little bit dodgy, I believe replace should be the behavior and should basically never happen in prod. I just encountered that in a unit test as we cannot really remove this value, at the same time I don't want to put an Option that will always be tested for a behavior that should never happen lol

cu29_log_runtime/src/lib.rs
mik90: Yeah I can make it into a separate benchmark in a follow up PR

PR #18, opened 2024-08-14
cu29/config: Added accumulator and store parameters for Connections

gbin:


cu29/src/config.rs
```diff
@@ -361,27 +372,35 @@ impl CuConfig {
...
-            })
-            .collect()
-    }
-
     /// Adds an edge between two nodes/tasks in the configuration graph.
     /// msg_type is the type of message exchanged between the two nodes/tasks.
-    pub fn connect(&mut self, source: NodeId, target: NodeId, msg_type: &str) {
-        self.graph
-            .add_edge(source.into(), target.into(), msg_type.to_string());
+    pub fn connect(
```
mik90, line 377: Considering users are rarely going to batch, maybe there should be a split between `connect` and `connect_batched`

There's also `store` but i dont think this is worth building a whole configuration type just for this unless you picture it being a heavy call-site for users with more configurations on the way. Sadly we can't pass in a `Cnx` struct since the nodes are indices here and `Cnx` has them as strings

cu29/src/config.rs
```diff
@@ -227,15 +232,23 @@ pub struct Cnx {
...
     /// Message type exchanged betwee src and dst.
-    msg: String,
+    pub msg: String,
+
+    /// Tells Copper to batch messages before sending the buffer to the next node.
+    /// If None, Copper will just send 1 message at a time.
+    /// If Some(n), Copper will batch n messages before sending the buffer.
+    pub batch: Option<u32>,
+
+    /// Tells Copper if it needs to log the messages.
```
mik90, line 242: Does 'log' entail both serialization and writing to disk here?

mik90:
> For example if the user has an IMU at 1KHz but only wants to get a measure every second, it will just set the edge to accumulate = 1000 and its node/task will be called every second.

I figure accumulate was meant to be updated to batch, but I wouldn't call this accumulating or batching data, it would just be sampling it. If it was batching the data, the task would be exposed to all 1000 measurements in a given execution. Then the task would either coalesce that data into a smaller count of outputs or match the arity and publish 1000 outputs, making the batching continue downstream

mik90:
> What do you think about the concept?

I think the approach of filtering large amounts of data with micro-nodes is interesting. Normally I'd be concerned with overhead but the operations being performed are going to be performed anyway. Stuff like this needs to be done in order to simplify the actual task implementation.

The correlation between message count and timing is a bit worrisome when it comes to resource availability. A sensor may generally provide 1KHz data but if we're going up in frequency and are doing many other things on the machine, we won't necessarily see steady data rates. That's not really something you can necessarily do much about, but you could implement temporal triggering. Granted, if you do that, then any guarantees on translating N values into Y values goes out the window because it's all based on resource availability and that's not particularly better

cu29/src/config.rs
mik90: Actually, maybe `CnxConfig` would be fine. It'd be stored in a `Cnx` and then you can get some nice `Default()` semantics while making it

mik90:
Maybe these micro nodes are how you'd implement 'combiners' that connect outputs from some nodes A and B which get aggregated into an input to node C

cu29/src/config.rs
gbin: yes.
Also we need to get to vocabulary right, I am so bad at this :).

gbin:
For the perf:

For the upstream performance, I think if really this is an issue the driver can always send buffers instead of one by one samples but my gut feeling is that it wont be necessary as if the sampler is right after the driver the compiler should see through this and inline the buffering basically.

> The correlation between message count and timing is a bit worrisome when it comes to resource availability. A sensor may generally provide 1KHz data but if we're going up in frequency and are doing many other things on the machine, we won't necessarily see steady data rates. That's not really something you can necessarily do much about, but you could implement temporal triggering. Granted, if you do that, then any guarantees on translating N values into Y values goes out the window because it's all based on resource availability and that's not particularly better

 This is a good one, the next big topic is basically "time alignment" in that category. At some point we need to be able to have a feature that for a given reference measure, take the closest from this set of sample from this other source.

For example: you need the attitude of the robot at the time of a video frame: you take a buffer from the IMU/AHRS and pick the closest one from the time of validity of the video frame. 
This could be implemented as an extension of this PR with a triggering source from anything upstream in the copperlist?

gbin:
> Maybe these micro nodes are how you'd implement 'combiners' that connect outputs from some nodes A and B which get aggregated into an input to node C

Ha you read my mind, this is the same concept of the "aligner" I am talking about just in the previous comment!

gbin:
> The correlation between message count and timing is a bit worrisome when it comes to resource availability. A sensor may generally provide 1KHz data but if we're going up in frequency and are doing many other things on the machine,

Yes. Games usually fix that with some dynamic "slop" if the system is late to avoid skipping a full frame/cycle + a deadline on the other end if really the data is too late It is like a dynamic window of time that is trying to catch the events around where there are expected. I almost added those parameters too but maybe we should go incrementally building up to this.

cu29/src/config.rs
gbin: This API was made because I wanted to use PetGraph but really did not want to expose it to the user as it is way more general than what copper needs (but its features for searching nodes etc are pretty good for the engine itself)

Let me split the 2 methods and rename that to sampler for now. Maybe we can think of a kind of builder pattern to simplify the API at some point, it is usually a good way to add "optionality" in Rust APIs it seems like

PR #21, opened 2024-08-16
runtime loops: Part 1: added a notion of loop for the runtime plan

gbin:
on the back of adding batches, this adds the notion of loops in the runtime plan.

### DIFF
--- a/cu29/src/curuntime.rs
+++ b/cu29/src/curuntime.rs
@@ -120,10 +120,33 @@ pub struct CuExecutionStep {
     pub culist_output_index: Option<u32>,
 }
 
-fn find_output_index_from_nodeid(node_id: NodeId, steps: &Vec<CuExecutionStep>) -> Option<u32> {
+/// This structure represents a loop in the execution plan.
+/// It is used to represent a sequence of steps that are executed multiple times.
+/// if loop_count is None, the loop is infinite.
+pub struct CuExecutionLoop {
+    pub steps: Vec<CuExecutionUnit>,
+    pub loop_count: Option<u32>,
+}
+
+/// This structure represents a step in the execution plan.
+pub enum CuExecutionUnit {
+    Step(CuExecutionStep),
+    Loop(CuExecutionLoop),
+}
+
+fn find_output_index_from_nodeid(node_id: NodeId, steps: &Vec<CuExecutionUnit>) -> Option<u32> {
     for step in steps {
-        if step.node_id == node_id {
-            return step.culist_output_index;
+        match step {
+            CuExecutionUnit::Loop(loop_unit) => {
+                if let Some(index) = find_output_index_from_nodeid(node_id, &loop_unit.steps) {
+                    return Some(index);
+                }
+            }
+            CuExecutionUnit::Step(step) => {
+                if step.node_id == node_id {
+                    return step.culist_output_index;
+                }
+            }
         }
     }
     None
@@ -131,13 +154,13 @@ fn find_output_index_from_nodeid(node_id: NodeId, steps: &Vec<CuExecutionStep>) 
 
 /// This is the main heuristics to compute an execution plan at compilation time.
 /// TODO: Make that heuristic plugable.
-pub fn compute_runtime_plan(config: &CuConfig) -> CuResult<Vec<CuExecutionStep>> {
+pub fn compute_runtime_plan(config: &CuConfig) -> CuResult<CuExecutionLoop> {
     let mut next_culist_output_index = 0u32;
 
     // prob not exactly what we want but to get us started
     let mut visitor = Bfs::new(&config.graph, 0.into());
 
-    let mut result = Vec::new();
+    let mut result: Vec<CuExecutionUnit> = Vec::new();
 
     while let Some(node) = visitor.next(&config.graph) {
         let id = node.index() as NodeId;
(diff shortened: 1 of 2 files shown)


cu29/src/curuntime.rs
```diff
@@ -217,9 +240,12 @@ pub fn compute_runtime_plan(config: &CuConfig) -> CuResult<Vec<CuExecutionStep>>
             output_msg_type,
             culist_output_index,
         };
-        result.push(step);
+        result.push(CuExecutionUnit::Step(step));
     }
-    Ok(result)
+    Ok(CuExecutionLoop {
+        steps: result,
+        loop_count: None, // if in not unit testing, the main loop is infinite
```
mik90, line 247: It'd be helpful to have a of node that can cause the execution to early-exit. 

This is useful for tests, simulations, etc that are meant to exit on a given condition in either a clean exit (return code 0) or an error exit (return code non-zero).

This would allow you to remove `loop_count` entirely and only allow the execution to stop on a specific condition or with Ctrl+C

cu29_derive/src/lib.rs
```diff
@@ -397,15 +401,19 @@ fn extract_tasks_types(copper_config: &CuConfig) -> (Vec<String>, Vec<Type>) {
...
-                None
-            } else {
-                Some(parse_str::<Type>(step.output_msg_type.clone().unwrap().as_str()).unwrap())
+        .filter_map(|unit| match unit {
+            CuExecutionUnit::Step(step) => {
+                if step.output_msg_type.is_none() {
+                    None
+                } else {
+                    Some(parse_str::<Type>(step.output_msg_type.clone().unwrap().as_str()).unwrap())
+                }
```
mik90, line 414: we can use either `if let` or `match` here instead of using unwrap on `step.output_msg_type`

cu29_derive/src/lib.rs
```diff
@@ -142,84 +142,88 @@ pub fn copper_runtime(args: TokenStream, input: TokenStream) -> TokenStream {
...
-                }
-            };
+                        CuTaskType::Regular => {
+                            let input_culist_index = int2index(
+                                step.culist_input_index
+                                    .expect("Sink task should have an input message index."),
+                            );
+                            let output_culist_index = int2index(
+                                step.culist_output_index
+                                    .expect("Src task should have an output message index."),
```
mik90, line 207: minor, but i noticed that these error messages refer to src/sink tasks and not Regular

```suggestion
                                    .expect("Regular task should have an input message index."),
                            );
                            let output_culist_index = int2index(
                                step.culist_output_index
                                    .expect("Regular task should have an output message index."),
```

cu29/src/curuntime.rs
```diff
@@ -120,24 +120,47 @@ pub struct CuExecutionStep {
     pub culist_output_index: Option<u32>,
 }
 
-fn find_output_index_from_nodeid(node_id: NodeId, steps: &Vec<CuExecutionStep>) -> Option<u32> {
+/// This structure represents a loop in the execution plan.
+/// It is used to represent a sequence of steps that are executed multiple times.
```
mik90, line 124: It's not as simple as a sequence of steps since it represents either sequences of steps or sequences _of_ sequences of steps, right?

cu29_derive/src/lib.rs
```diff
@@ -397,15 +401,19 @@ fn extract_tasks_types(copper_config: &CuConfig) -> (Vec<String>, Vec<Type>) {
...
-                Some(parse_str::<Type>(step.output_msg_type.clone().unwrap().as_str()).unwrap())
+        .filter_map(|unit| match unit {
+            CuExecutionUnit::Step(step) => {
+                if step.output_msg_type.is_none() {
+                    None
+                } else {
+                    Some(parse_str::<Type>(step.output_msg_type.clone().unwrap().as_str()).unwrap())
+                }
             }
+            CuExecutionUnit::Loop(_) => todo!("Needs to be implemented"),
```
mik90, line 416: Is it going to go through all of the steps in the loop and aggregate all of the message types? 

And do we care about duplicates in this return type? Should it be changed into a set or is the type vec being indexed into or something?

cu29/src/curuntime.rs
gbin: but here we are at compile time, this is just to represent the overall structure to execute, not really to execute it itself.

But I get what you mean, we could add a "debug" feature that can at runtime override the exit of the plan to do a kind of conditional breakpoint of some sort. Let's add that to the todo list.

cu29_derive/src/lib.rs
gbin: oh yeah, this is ugly let me fix this one :) Thanks

cu29_derive/src/lib.rs
gbin: oh nicely spotted, fixing that, it would have been confusing for the user :)

cu29/src/curuntime.rs
gbin: yes, fixed.

cu29_derive/src/lib.rs
gbin: I believe it works as is ...  we will need an array of the type instead of just a type for each step of the loop as output. And at runtime we need to generate the index for the loop.
in the vector it will be:
[mainloop_type1 [loop1_type1[] loop1type2[]] mainloop_type2 ...] 

we might want to go through some design lol

PR #23, opened 2024-08-22
unifiedlog: refactor for slabs.

gbin:
This is to prepare for the next PR for a multislab version of the unified logger


gbin:
This was merged with the previous PR

PR #25, opened 2024-08-23
Adds macos to the CI/CD pipeline

gbin:


.github/workflows/general.yml
```diff
@@ -11,7 +11,11 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-13 ]
```
mik90, line 18: Ya id recommend leaving both hard coded to major/minor versions

Maybe have another step with latest to test it out but just for the sake of knowing a given release is working and not getting surprised 

Moreso for mac than Ubuntu but it's useful to do for both

gbin: done

PR #42, opened 2024-09-13
Fix cu29 value bincode deserialization

kamibo:


gbin:
Thanks @kamibo  That was b0rken for sure :) for ref, this has to match Value::discriminant.

gbin (approved):

PR #44, opened 2024-09-19
API refactoring to allow not only multiple inputs but also optional inputs and outputs.

gbin:


gbin:
As the API is changing the templated project generation is broken (it picks up the master branch)

cu29/src/curuntime.rs
```diff
@@ -129,122 +148,221 @@ pub struct CuExecutionLoop {
...
+            } else {
+                false
+            }
+        }) {
+            // modify the existing step and put it back in the plan as the current step as it needs this subsequent output.
+            let mut step = plan.remove(pos);
+            if let CuExecutionUnit::Step(ref mut s) = step {
+                s.input_msg_indices_types = input_msg_indices_types;
+            }
+            plan.push(step);
```
mik90, line 320: Why not modify the step in-place instead of popping and pushing it?

cu29/src/cutask.rs
```diff
@@ -16,8 +16,51 @@ use std::fmt::{Display, Formatter};
...
+
+impl<'cl, T: CuMsgPayload> CuMsgPack<'cl> for (&'cl CuMsg<T>,) {}
+impl<'cl, T: CuMsgPayload> CuMsgPack<'cl> for &'cl CuMsg<T> {}
+impl<'cl, T: CuMsgPayload> CuMsgPack<'cl> for (&'cl mut CuMsg<T>,) {}
+impl<'cl, T: CuMsgPayload> CuMsgPack<'cl> for &'cl mut CuMsg<T> {}
+impl<'cl> CuMsgPack<'cl> for () {}
+
+// Apply the macro to generate implementations for tuple sizes up to 5
+impl_cu_msg_pack! {
+    (T1, T2), (T1, T2, T3), (T1, T2, T3, T4), (T1, T2, T3, T4, T5) // TODO: continue if necessary
```
mik90, line 40: Can we automate this with something like https://docs.rs/variadics/latest/variadics/#var_type?

templates/cu_full/src/tasks.rs
```diff
@@ -66,18 +62,18 @@ impl CuTaskLifecycle for MyTask {
     // don't forget the other lifecycle methods if you need them: start, stop, preprocess, postprocess
 }
 
-impl CuTask for MyTask {
-    type Input = MyMsg;
-    type Output = MyMsg;
+impl<'cl> CuTask<'cl> for MyTask {
+    type Input = input_msg!(MyPayload);
```
mik90, line 66: In order to remove the optional from user-space, im wondering if we'd be able to have the input msg like a macro as a struct, so

```rust
#[cu_input]
struct MyPayload {
  my_optional_value: Option<i32>,
  my_required_value: f32,
  my_array: Vec<u32>,
}
```

and the runtime would desugar these into payloads with optional values but only execute a CuTask if all its required inputs are available during runtime. This may better relate to optional outputs where a given node may or may not produce a value depending on the content of its inputs

examples/cu_caterpillar/.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[build]
+rustflags = ["-Z", "macro-backtrace"]
```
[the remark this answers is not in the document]
gbin: ok removed, it only works on nightly, this helps debugging the macros :)

examples/cu_multisources/src/tasks.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl CuTaskLifecycle for StdTask {
+    fn new(_config: Option<&NodeInstanceConfig>) -> CuResult<Self> {
+        Ok(Self {})
+    }
+}
+
+impl<'cl> CuTask<'cl> for StdTask {
+    type Input = input_msg!('cl, i32, f32);
+    type Output = output_msg!('cl, (i32, f32));
```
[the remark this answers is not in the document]
gbin: good point, let me add some comment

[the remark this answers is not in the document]
gbin: and change the naming

cu29/src/curuntime.rs
```diff
@@ -129,122 +148,221 @@ pub struct CuExecutionLoop {
...
-            output_msg_type = Some(
-                config
+        let mut input_msg_indices_types: Vec<(u32, String)> = Vec::new();
+        let mut output_msg_index_type: Option<(u32, String)> = None;
+
+        let task_type = find_task_type_for_id(&config.graph, id);
+
+        match task_type {
+            CuTaskType::Source => {
+                eprintln!("Visiting Source node: {:?}", id);
```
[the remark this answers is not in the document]
gbin: yep done, this was a bunch of debug info thanks

cu29/src/curuntime.rs
gbin: oh because you need to put it at the end of the copper list to be sure all the tasks sending it messages has been called prior to it.

cu29/src/cutask.rs
gbin: I looked into it a little bit but those recursive constructs would be exposed to the users I believe: instead of a tuple of &CuMsg you have a (&CuMsg, (&CuMsg, )) so you need to also have the user use the macros from variadics to unpack them.

I would say, let's wait that it becomes standard in Rust so we don't paint ourselves in a corner.

templates/cu_full/src/tasks.rs
gbin: This is an interesting idea!
Basically we would add an option in the tree to say: all inputs mandatory
Then the runtime can basically serve a straight struct with no Option<>

But we would need to keep the non mandatory one, for example if you have a dual sensor path in your stack and one goes dark and doesn't produce any data, roboticians will want a way to continue in some degraded mode.

The other one is a super high frequency sensor merging with a slow one, you need to continue to feed the merger task.

Let's merge it like that and think about it again before the interface freeze at the beta stage. Let me add an issue to be sure we address it.

Thanks!

gbin:
Let me merge this to be sure the project generation works.

PR #46, opened 2024-09-21
Balance bot demo implementation.

gbin:
This is WIP with some good refactoring in the drivers so it is worth it to merge it.


gbin:
Merging this state that is a good tuning point

PR #50, opened 2024-09-27
│First implementation of cumulative stats

gbin:
It allows to build real time statistics with a specialization for CuDurations out of Copper's engine.

### DIFF
--- a/cu29/src/monitoring.rs
+++ b/cu29/src/monitoring.rs
@@ -3,7 +3,7 @@
 
 use crate::config::ComponentConfig;
 use crate::cutask::CuMsgMetadata;
-use cu29_clock::RobotClock;
+use cu29_clock::{CuDuration, RobotClock};
 use cu29_traits::{CuError, CuResult};
 use std::alloc::{GlobalAlloc, Layout, System};
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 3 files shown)


cu29/src/monitoring.rs
```diff
@@ -122,3 +122,203 @@ impl Drop for ScopedAllocCounter {
...
+
+    #[inline]
+    pub fn percentile(&self, percentile: f64) -> u64 {
+        self.stats.value_at_quantile(percentile)
+    }
+
+    /// Adds a value to the statistics.
+    #[inline]
+    pub fn record(&mut self, value: u64) {
+        self.stats.record(value).unwrap();
```
mik90, line 169: Does this panic if we hit the max record count?

Not sure if that's ideal, it could only keep a running average and max/min in that case

mik90 (approved):
Looks like a good base for tracking a value in abstract, having configurable bounds is good. It's great to have toggles for tolerable overhead

cu29/src/monitoring.rs
gbin: indeed, that was a little lazy on my part this one :P

cu29/src/monitoring.rs
```diff
@@ -122,3 +122,203 @@ impl Drop for ScopedAllocCounter {
...
+
+use hdrhistogram::Histogram;
+
+/// Accumulative stat object that can give your some real time statistics.
+pub struct LiveStatistics {
+    stats: Histogram<u64>, // u64 is the Counter type.
+}
+
+impl LiveStatistics {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
gbin: indeed, we should be very explicit with reallocs

PR #51, opened 2024-09-27
Monitor part 2: the backend

gbin:
it is tested with the cu-monitor project


gbin:
Also sneaked in some renaming on the backend as we start to standardize around names: payload is for the messages, _instances are redundant and generation oriented etc...

PR #53, opened 2024-09-29
Sink virtual output + Added wrapper around msgs instead of tuple.

gbin:
It solves 3 problems:
1. now we can track the latency truely end to end without the hack I implemented previously forcing me to have the input for the sinks as writeable. Now it generates a virtual end msg that is just the metadata for the sink.
2. it allows monitoring to work with sinks
3. doing that I hit a serious limitation with the tuple that triggered a refactoring of the CopperLists. It is 2 fold: one, all the blanket traits like Debug are implemented up to tuples of 12 elements, and the second issue, any tuple is considered a foreign type so if the trait is not in the crate I cannot implement them. The solution is to have a type wrapper.

While at it, I renames cl.payload to cl.msgs so it makes the hierarcy crystal clear:
CL - Msg - Payload

### DIFF
--- a/cu29/src/copperlist.rs
+++ b/cu29/src/copperlist.rs
@@ -47,16 +47,16 @@ impl Display for CopperListState {
 pub struct CopperList<P: CopperListTuple> {
     pub id: u32,
     state: CopperListState,
-    pub payload: P, // This is generated from the runtime.
+    pub msgs: P, // This is generated from the runtime.
 }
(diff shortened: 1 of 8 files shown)


mik90 (approved):

gbin:
alright the templating example fails because I changed stuff not reflected on master. we should be good to go on the reader side too now.

PR #55, opened 2024-10-03
Monitoring UI

gbin:
It is a small TUI implemented from a Monitoring component for Copper.

A quick sys info 
image

The error state of all the tasks in the dag
image

All the latencies stats, including end to end:
image


examples/cu_consolemon/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
```
mik90, line 1: Should this live in examples? Seems like it'd fit better in a tools/ directory

examples/cu_caterpillar/copperconfig.ron
```diff
@@ -113,4 +113,8 @@
         (src: "ct-6", dst: "gpio-6", msg: "cu_rp_gpio::RPGpioPayload"),
         (src: "ct-7", dst: "gpio-7", msg: "cu_rp_gpio::RPGpioPayload"),
     ],
+    monitor: (
```
mik90, line 116: This section means that 'cargo run' actually runs the monitor in the terminal, it'd be worth noting that here

Maybe it'd be worth handling as a CLI param e.g. `cargo r -- --monitor` or the opposite for opting out like `cargo r -- --no-monitor` /  `cargo r -- --headless`

examples/cu_consolemon/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+                        KeyCode::Char('2') => self.active_screen = Screen::Dag,
+                        KeyCode::Char('3') => self.active_screen = Screen::Latency,
+                        KeyCode::Char('r') => {
+                            if self.active_screen == Screen::Latency {
+                                self.task_stats.lock().unwrap().reset()
+                            }
+                        }
+                        KeyCode::Char('q') => {
+                            // Basically triggers the clean ctrlc logic of copper.
+                            signal::kill(
```
mik90, line 281: I don't think this is resetting the terminal correctly. When i do `q` it exits but dumps out a bunch of ansi color commands

Screenshot 2024-10-02 210506

I have to shut down the terminal to stop it from spitting out those strings endlessly.

I'm running Ubuntu 24.04 using WSL2

examples/cu_consolemon/src/sysinfo.rs
```diff
@@ -0,0 +1,220 @@
...
+        PfetchInfo::Wm => pfetch::wm(&readouts.general_readout),
+        PfetchInfo::De => pfetch::de(&readouts.general_readout),
+        PfetchInfo::Palette => Some("Color Palette".to_string()), // Simplified palette display
+        PfetchInfo::BlankLine => Some("".to_string()),
+    }
+}
+
+// Function to render the gathered info along with ASCII logo
+pub fn pfetch_info() -> String {
+    let readouts = Readouts {
```
zachgoins, line 89: Nit: Probably worth just making a `::default()` method on `Readouts` so you can just call `Readouts::default()` wherever you want the default initialization.

examples/cu_consolemon/src/sysinfo.rs
```diff
@@ -0,0 +1,220 @@
...
+    BlankLine,
+}
+
+impl Display for PfetchInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", format!("{self:?}").to_lowercase())
+    }
+}
+
+impl FromStr for PfetchInfo {
```
zachgoins, line 33: The [https://docs.rs/strum/latest/strum/](strum) crate has a lot of this "enum to string or string to enum" behavior built in. It's pretty light and probably worth bringing it in as a dependency just for it's usefulness there. Like you can just put a `#[derive(strum::AsRefStr)]` above the enum you want string conversions for and it will generate all the conversions for you. Check it out!

examples/cu_consolemon/src/sysinfo.rs
```diff
@@ -0,0 +1,220 @@
...
+        PfetchInfo::De,
+    ];
+
+    let logo = pfetch::logo(&os);
+    let gathered_pfetch_info: Vec<(Color, String, String)> = all_infos
+        .iter()
+        .filter_map(|info| match info {
+            PfetchInfo::Os => Some((logo.primary_color, info.to_string(), os.clone())),
+            _ => get_info(&info, &readouts).map(|info_str| match info {
+                PfetchInfo::Title => (logo.secondary_color, info_str, "".to_string()),
```
zachgoins, line 117: MEGA nit: You can use `.into()` instead of `.to_string()` in a lot of places. Definitely a preference thing. But I like it just because it keep things a little more concise.

zachgoins (commented):
I am super out of the loop on things so I didn't check the behavior. But I did leave a few little rustfoo nit comments. Hope they are helpful! Leaving it to others more in the loop to approve though

examples/cu_consolemon/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+        TaskStats {
+            stats,
+            end2end: CuDurationStatistics::new(max_duration),
+        }
+    }
+
+    fn update(&mut self, msgs: &[&CuMsgMetadata]) {
+        for (i, &msg) in msgs.iter().enumerate() {
+            let (before, after) = (msg.before_process.unwrap(), msg.after_process.unwrap());
+            self.stats[i].record(after - before);
```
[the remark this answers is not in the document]
gbin: So the messages are always here. the payload of the messages are optional. So it should never have a different length and they always map 1 to 1.

examples/cu_consolemon/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn reset(&mut self) {
+        for s in &mut self.stats {
+            s.reset();
+        }
+        self.end2end.reset();
+    }
+}
+
+fn dag(config: &CuConfig) -> String {
+    "test".to_string()
```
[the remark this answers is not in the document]
gbin: ha.. got implemented, it was just a vestigial placeholder sorry.

examples/cu_consolemon/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn run_app<B: Backend>(&mut self, terminal: &mut Terminal<B>) -> io::Result<()> {
+        loop {
+            terminal.draw(|f| {
+                self.draw(f);
+            })?;
+
+            if event::poll(Duration::from_millis(10))? {
+                if let Event::Key(key) = event::read()? {
+                    match key.code {
+                        KeyCode::Char('1') => self.active_screen = Screen::Neofetch,
```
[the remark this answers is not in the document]
gbin: I did reuse a rust implementation of a neofetch clone :)

examples/cu_consolemon/Cargo.toml
gbin: good call I think we need to reorg the repo. Also, now that I cleaned up the PID algorithm, I think we can have a little tree of drivers/algorithms/monitors as proper components. I will do that after merging this because another big PR is based upon this one.

examples/cu_caterpillar/copperconfig.ron
gbin: Very good point but I think this is more the exception than the rule here as I am aiming at a good demo of this component. Normal monitors should be headless. 
What I am planning to do is to have some composable configs at some points where should be able to pick subsets of components dependendong on the specific robot you are deploying it on. I think we should put it in that logic, I am adding an issue so we can remember that.

examples/cu_consolemon/src/lib.rs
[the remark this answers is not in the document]
mik90: Ah gotcha, it'd be worth noting that constraint as a comment. This indexing is definitely more lightweight than doing a key value lookup

examples/cu_consolemon/src/lib.rs
gbin: Ok this threw me into a refactor, there were other shudown issues in the core too :). Now I send an error to the runtime on the process copperlist that makes the runtime shutting down correctly (and there were gremlins in the unified logger that was not closing properly that got fixed too)

examples/cu_consolemon/src/sysinfo.rs
gbin: I needed to copy paste some entry point from this library using libmacchina and now I regret it lol, let me fix this thanks.

PR #64, opened 2024-10-14
[Docker] Update 'run_docker.sh' to mount the entire workspace directory

Massi0:
It appears that the Docker mount path required an update following the [reorganization](https://github.com/copper-project/copper-rs/commit/55d02fd26063e59f5fc14792cf2a93b4ca221e39).

Tested locally.


gbin:
Indeed thanks!

PR #72, opened 2024-10-24
disable the rp specific depencies on non linux platforms

gbin:


gbin:
@Massi0 Do you mind testing this crate to see if it compiles correctly in mock mode on MacOS?

Thanks!

gbin:
alright let's merge this, I'll try to have --workspace compile on CI/CD so mac won't regress on sim

PR #73, opened 2024-10-24
Fix: terminal not properly restored on exit + strerr bombing prevention.

gbin:
The terminal issues became a huge pain with the simulators. This should be fixing the corruptions from stderr and retoring the terminal.


mik90:
Just pulled this down and tested it out on my laptop running fedora + bash in the vscode integrated terminal. Exit works well now!

I did test out `killall cu-caterpillar` (should default to SIGTERM)  out of sheer curiosity and rudeness, and it didn't handle that well. It just prints a bunch of ANSI color codes. Not important tho, this is already an improvement

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -585,9 +577,29 @@ impl CuMonitor for CuConsoleMon {
...
+            let backend = CrosstermBackend::new(stdout());
+
+            setup_terminal();
+
+            // nukes stderr into orbit as anything in the stack can corrupt the terminal
+            let dev_null = File::open("/dev/null").expect("Could not get /dev/null");
+            let dev_null_fd = dev_null.as_raw_fd();
+            let original_stderr_fd = unsafe { dup(2) };
+            unsafe {
+                dup2(dev_null_fd, 2);
```
mik90, line 589: Does the libc create have a constant for stderr or do you have to use `2`? If not, could be good to use a var to denote that. I had to double check https://en.wikipedia.org/wiki/File_descriptor

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -664,7 +669,17 @@ fn init_error_hooks() {
     }));
 }
 
+fn setup_terminal() {
+    enable_raw_mode()
+        .map_err(|e| CuError::new_with_cause("Could not enable console raw mode", e))
+        .expect("Could not enter raw mode: check terminal compatibility.");
```
mik90, line 675: Do we need both map_err and expect here? I think the latter would do

mik90 (approved):

components/monitors/cu_consolemon/src/lib.rs
gbin: oh yeah, indeed a brainless copy paste / autocomplete line lol

components/monitors/cu_consolemon/src/lib.rs
gbin: ha yes, libc::STDERR_FILENO  it is

PR #86, opened 2024-11-12
fix reset_sim of cu_rp_balancebot

makeecat:
The y and z component of pole transform need to be reset, it help recovering the pole when it is below the cart. (it is easier to recover to pause then reset)

It would be better to reset the velocity, but seems like we have no access to that (it is handled by the physics engine).


gbin:
?Thanks

PR #101, opened 2024-11-15
The Copper side for the Hesai driver

gbin:


components/sources/cu_hesai/src/lib.rs
```diff
@@ -1 +1,263 @@
...
+        sysctl::CtlValue::Uint(value) => Some(value),
+        sysctl::CtlValue::String(value) => value.parse::<u32>().ok(),
+        other => {
+            println!("Unexpected sysctl value type: {:?}", other);
+            None
+        }
+    }
+}
+
+fn spherical_to_cartesian(
```
luc-k1, line 34: Could you specify angle convention used for spherical coordinates

components/sources/cu_hesai/src/lib.rs
```diff
@@ -1 +1,263 @@
...
+            Ok(size) => {
+                let lidar_packet = parser::parse_packet(&buf[..size])
+                    .map_err(|e| CuError::new_with_cause("Failed to parse Hesai UDP packet", e))?;
+
+                // this is the reference point for the block timings
+                let t6 = lidar_packet
+                    .block_ts(&self.reftime)
+                    .map_err(|e| CuError::new_with_cause("Failed to get block timings", e))?[5]; // 0 == channel 1, 5 == channel 6
+
+                let is_dual = lidar_packet.header.is_dual_return();
```
luc-k1, line 125: Are we doing something this this information ?
How would we handle multiple returns in LidarCuMsgPayload ?

luc-k1 (approved):
Main concern would be for LidarPayloadSoa and handling of other lidar providing more information (multiple return 2+, or dual channel)

components/sources/cu_hesai/src/lib.rs
gbin: ok let's drop that for that PR and we will add another one with a proper return order in the format and the drivers. thanks

PR #106, opened 2024-11-19
Introducing CuTimeRange

gbin:
This is needed to represent messages that contains more than 1 Tov like several images, several IMU measures etc... We will add helpers like range merging, range intersections etc in a separate PRs.

### DIFF
--- a/components/sources/cu_hesai/src/lib.rs
+++ b/components/sources/cu_hesai/src/lib.rs
@@ -7,7 +7,7 @@ use cu29::cutask::CuMsg;
 use cu29::cutask::{CuSrcTask, CuTaskLifecycle, Freezable};
 use cu29::{output_msg, CuError, CuResult};
 use cu29_clock::{CuDuration, CuTime, CuTimeRange, RobotClock, Tov};
-use cu_sensor_payloads::{LidarPayload, LidarPayloadSoa};
+use cu_sensor_payloads::{PointCloud, PointCloudSoa};
 use socket2::{Domain, Protocol, SockAddr, Socket, Type};
 use std::io::ErrorKind;
 use std::io::Read;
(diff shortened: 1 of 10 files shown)


components/sources/cu_hesai/src/lib.rs
```diff
@@ -121,7 +121,12 @@ impl<'cl> CuSrcTask<'cl> for Xt32 {
                         payload.push(LidarPayload::new_uom(t, x, y, z, r, None));
                     });
                 }
-                new_msg.metadata.tov = payload.tov[0].into(); // take the oldest timestamp
+
+                let tov_range = CuTimeRange {
+                    start: payload.tov[0],
```
mik90, line 126: Since this is a soa, can we build a CuTimeRange from a slice?

So impl From or TryFrom for CuTimeRange and get `CuTimeRange::from(&[T])`

Syntactic sugar but still useful

components/tasks/cu_pid/src/lib.rs
```diff
@@ -229,7 +229,11 @@ where
     ) -> CuResult<()> {
         match input.payload() {
             Some(payload) => {
-                let tov = input.metadata.tov.unwrap();
+                let tov = match input.metadata.tov {
+                    Tov::Time(single) => single,
+                    _ => panic!("Unexpected variant"),
```
mik90, line 234: Is this panicking because nothing should ever continue if this happens?

An early-return with Err would be more configurable for error handling

mik90 (approved):

core/cu29_clock/src/lib.rs
```diff
@@ -233,6 +233,59 @@ impl From<CuTime> for OptionCuTime {
     }
 }
 
+/// Represents a time range.
+#[derive(Clone, Debug, Encode, Decode, Serialize, Deserialize)]
+pub struct CuTimeRange {
+    pub start: CuTime,
```
mik90, line 239: `CuTime` is `Copy`, `CuTimeRange` can be copy too can't it https://doc.rust-lang.org/std/marker/trait.Copy.html#when-can-my-type-be-copy?

This would remove the need to do `clone()` in the task code

components/sources/cu_hesai/src/lib.rs
gbin: Added it. It is not used yet but it is a nice thing to have.

components/tasks/cu_pid/src/lib.rs
gbin: indeed, fixed thanks

core/cu29_clock/src/lib.rs
gbin: good point, fixed.

PR #115, opened 2024-11-28
Removes the lifecycle trait

gbin:
It split artificially the implementation of tasks and couples the passed types.


mik90 (approved):
Flattening the hierarchy makes sense, this makes it easier to read

PR #127, opened 2024-12-04
Dependency cleaning / Better Windows compat

gbin:


mik90:
If you did these manually, there's likely tooling that'll automate this. I found https://github.com/est31/cargo-udeps

Feel free to ignore if you were already using it ofv

gbin:
> If you did these manually, there's likely tooling that'll automate this. I found https://github.com/est31/cargo-udeps
> 
> Feel free to ignore if you were already using it ofc

Ha yes, I was using rustrover that just release the feature in the IDE but it is not perfect: some crates contains macros and those macros generated code depends on some crate... Do you know if there is a better way? should we add the dep to the macro crate and reshare the deps or something? @mik90

mik90:
> > If you did these manually, there's likely tooling that'll automate this. I found https://github.com/est31/cargo-udeps
> > 
> > Feel free to ignore if you were already using it ofc
> 
> Ha yes, I was using rustrover that just release the feature in the IDE but it is not perfect: some crates contains macros and those macros generated code depends on some crate... Do you know if there is a better way? should we add the dep to the macro crate and reshare the deps or something? @mik90   

Not sure, it's possible Jetbrains is just using some open source dependency linter like udeps  but it's also possible they rolled their own

Idk what's best practice, but I imagine that if the macro always generates code that relies on a dep, it's best to make that dep a dep of the macro so at least the version used is documented and versions aren't mixed. I'm not talking from experience though, might be better to look on the rust forums or rust discord.

PR #130, opened 2024-12-05
Move setup_entities to Startup phase

Nashenas88:
Running setup_entities during the Update phase was causing major performance issues. It should only need to be called once. Otherwise the example will hit major frame rate and memory issues.

On my system, the example went from <1 fps to ~45.


gbin:
oh that was a dumb mistake! Thank you for spotting it and fixing it!

PR #137, opened 2024-12-05
Bug fixes and cleanups for balancebot sim

gbin:


gbin:
@Nashenas88  actually it was needed :)...  because we have a race condition spawning the entities vs the resources loading. I added a boolean flag not not recreate the entities once they are build.

Do you mind checking if that restores the fps on your side too? Thanks

gbin:
> @Nashenas88 actually it was needed :)... 

I consistently get 60fps+ on my small integrated graphics laptop. :+1:

Nashenas88:
The current version works. Thanks!

PR #145, opened 2024-12-11
Fix and refactoring of PcapStreamer

gbin:
Only one survived.

Improved it by passing the expected packet payload size to send.


components/sources/cu_hesai/src/lib.rs
```diff
@@ -235,7 +165,8 @@ mod tests {
 
         xt32.reftime = (datetime, clock.now());
 
-        while streamer.send_next() {
+        // 1076 is the expected payload size for Hesai XT32
+        while streamer.send_next::<1076>() {
```
luc-k1, line 169: Not a fan of magic static value, since its statically deductible from packet struct  with a size_of::<lidar_parser::Packet>

components/sources/cu_vlp16/src/lib.rs
```diff
@@ -94,30 +94,23 @@ impl<'cl> CuSrcTask<'cl> for Vlp16 {
...
 
         // Read test.pcap
-        if let Some(pkt) = pcap_reader.next_packet() {
-            let pkt = pkt.unwrap();
-            let data = &pkt.data[0x2a..];
-            // send udp packet to 2368
-            let socket = UdpSocket::bind("0.0.0.0:2367").unwrap();
-            socket.send_to(data, "127.0.0.1:2368").unwrap();
-            // process
+        if streamer.send_next::<1206>() {
```
luc-k1, line 113: same remark

components/testing/cu_udpinject/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "cu-udpinject"
```
luc-k1, line 2: cu-udp-inject and cu_udp_inject in project would be better consistant with other packages (and easier to read)

components/testing/cu_udpinject/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+        let packet = match self.capture.next_packet() {
+            Ok(packet) => packet,
+            Err(_) => return false, // End of the stream
+        };
+
+        // Assume 42-byte header (Ethernet + IP + UDP) and an optional 4-byte FCS
+        let payload_offset = 42;
+
+        // Extract only the payload, excluding headers and trailing FCS if present
+        let payload = &packet.data[payload_offset..payload_offset + PS];
```
luc-k1, line 43: A test is missing : if the packet is smaller than expected, the slicing will panick (case occurs mix a traffic of mixed size packets)

components/testing/cu_udpinject/Cargo.toml
gbin: done

components/testing/cu_udpinject/src/lib.rs
gbin: done

gbin:
thanks for the review @luc-k1

PR #151, opened 2024-12-13
WIP: Small PoC for generating schema for messages?

gbin:


core/cu29_derive/src/schema.rs
```diff
@@ -0,0 +1,45 @@
...
+    let schema = match &input.data {
+        syn::Data::Struct(data) => {
+            let fields: Vec<_> = data
+                .fields
+                .iter()
+                .map(|f| {
+                    let field_name = f.ident.as_ref().unwrap().to_string();
+                    let field_type = f.ty.to_token_stream().to_string();
+                    quote! {
+                        schema.push((#field_name.to_string(), #field_type.to_string()));
```
eggyal, line 18: I think this could cause issues because the same `field_type` can be represented in multiple ways (eg via a type alias or path etc).  Better I think to define a trait:

```rust
pub trait Schema {
    fn schema() -> Vec<(String, String)>;
}
```

and then copper can provide a `derive` macro for ease of implementation, which would delegate to the respective `Schema::schema()` of each field's type (with implementations provided for primitives and library types).

I'm also not entirely convinced about `(String, String)`.  Better I think to have a richer type that can be reliably mapped onto any primitive types known to different schema formats.

eggyal (commented):
Thanks for knocking this demo together.  I can see a path forward, but I'm not convince that is 100% it.  Leave it with me!

core/cu29_derive/src/schema.rs
gbin: yes definitely the String->String is not what we want, this is just some quick stuff generated with chatGPT to illustrate the general direction. We should probably go with Value or other enum based stuff?

As for the trait, that is interesting because it will be directly available from the tuple which is pretty cool

gbin:
I will be closing this PR, it was just to share with you the scrappy PoC, feel free to continue the discussion on the Issue

PR #182, opened 2024-12-23
add image payload

Lishen1:


components/payloads/cu_sensor_payloads/src/lib.rs
```diff
@@ -106,6 +110,150 @@ impl PointCloud {
...
+impl CuImage {
+    pub fn new(img: DynamicImage) -> Self {
+        Self(img)
+    }
+}
+
+impl Encode for CuImage {
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        // println!("enc w: {} h: {}", self.width(), self.height());
+        Encode::encode(&self.width(), encoder)?;
```
gbin, line 170: Thanks for the PR, it helps settling on something!

I think we want all that (ie width, height, stride, pixel type) to be static, ie in the type itself Cuimage<1920, 1200, RGBA, 1920>.
Like that for the images in (main) memory we know their size (they implement Sized) and we can simply preallocate them like we do for the rest.
Now (see the discussion on discord) we might want also a GPU / NPU / NUMA type for images that are no in the CopperList.

gbin: For the first part, I think I am wrong: looking at some hw, the stride can change (I see some camera with dynamic cropping for example)

gbin:
Let's close this one and focus on the compression side now thank you so much!

PR #184, opened 2024-12-25
V4l2 source

gbin:


components/payloads/cu_sensor_payloads/Cargo.toml
```diff
@@ -17,4 +17,5 @@ cu29-clock = { workspace = true }
 cu29 = { workspace = true }
 uom = { workspace = true }
 derive_more = { workspace = true }
+image = "0.25.5"
```
edgarriba, line 20: maybe optional so that users we can decide which backend image to use. There several alternatives to this e.g the kornia [Image](https://github.com/kornia/kornia-rs/blob/main/crates/kornia-image/src/image.rs#L59); ndarray (that can be represented as image); same story for Tensor libraries

gbin: Good point!

PR #186, opened 2024-12-28
Logging parameters in the Ron file. Fixes #183

### LINKED ISSUE #183: Logger is crashing logging images
From @edgarriba

image

Trying: https://github.com/kornia/kornia-rs/pull/193/files

gbin:
One for to override the section size (if the copperlist references allocated stuff outside of it) and one for the slab size (that will be hooked up after the construction from the copper_ctx is merged)


gbin:
It is also solving #156

Issue #192, opened 2024-12-31
Add Kornia image support to CuImage


gbin:
kornia-rs compatibility would allow a whole range of CV algorithms to be used directly in Copper

gbin:
done

PR #193, opened 2024-12-31
Added Kornia image support

gbin:
Solving #192

### DIFF
--- a/components/payloads/cu_sensor_payloads/src/image.rs
+++ b/components/payloads/cu_sensor_payloads/src/image.rs
@@ -1,8 +1,15 @@
 use bincode::{Decode, Encode};
 use cu29::prelude::CuBufferHandle;
 
+#[allow(unused_imports)]
+use cu29::{CuError, CuResult};
+
 #[cfg(feature = "image")]
 use image::{ImageBuffer, Pixel};
+#[cfg(feature = "kornia")]
+use kornia::image::Image;
+#[cfg(feature = "kornia")]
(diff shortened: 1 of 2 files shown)


components/payloads/cu_sensor_payloads/src/image.rs
```diff
@@ -50,6 +57,28 @@ impl CuImage {
...
+        );
+
+        let size = width * height * C;
+
+        let raw_pixels: &[T] = unsafe {
+            core::slice::from_raw_parts(data.as_ptr() as *const T, data.len() / size_of::<T>())
+        };
+        let img_size: ImageSize = ImageSize { height, width };
+
+        unsafe { Image::from_raw_parts(img_size, raw_pixels.as_ptr(), size) }
```
edgarriba, line 81:
```suggestion
        unsafe { Image::from_raw_parts([width, height].into(), raw_pixels.as_ptr(), size) }
```
to save you from one feature line above

edgarriba: and why not passing directly `data.as_ptr()` ?

gbin: the buffers are u8 on Copper side and I need to pass data: *const T.

I initially went to the route of trying to have the element specified but even that would block a bunch of formats like NV12 where 12 is 12 bits per pixel Oo.. and biplanar :boom: 
Or bayered formats, or compressed buffers etc .. it is endless :). void * it will be for you for now lol.

PR #199, opened 2025-01-05
Make config entry to disable the logging in the ron file

### LINKED ISSUE #195: Make a config entry to disable the logging in the ron file
Some applications have very limited disk space so we might want to be able to disable the logging entirely from Copper.

AS1100K:
Fixes #195

### DIFF
--- a/core/cu29_runtime/src/config.rs
+++ b/core/cu29_runtime/src/config.rs
@@ -316,6 +316,8 @@ pub struct LoggingConfig {
     pub slab_size_mib: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub section_size_mib: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub disable_logging: Option<bool>
 }
 
 /// The config is a list of tasks and their connections.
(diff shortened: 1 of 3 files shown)


makeecat:
Let's wait till we finalize the discussion on #195

core/cu29_runtime/src/curuntime.rs
```diff
@@ -95,7 +106,9 @@ impl<CT, P: CopperListTuple + 'static, M: CuMonitor, const NBCL: usize> CuRuntim
             // serialize them all and Free them.
             if is_top && cl.get_state() == CopperListState::DoneProcessing {
                 cl.change_state(CopperListState::BeingSerialized);
```
gbin, line 108: should be in the if itself, it never get serialized :)

core/cu29_runtime/src/config.rs
```diff
@@ -705,12 +707,13 @@ mod tests {
     #[test]
     fn test_logging_parameters() {
         let txt =
-            r#"( tasks: [], cnx: [], logging: ( slab_size_mib: 1024, section_size_mib: 100, ),) "#;
+            r#"( tasks: [], cnx: [], logging: ( slab_size_mib: 1024, section_size_mib: 100, disable_logging: true ),) "#;
```
gbin, line 710: I think "logging" can be confusing as the other parameters relates to the unified logger (slab and sections) while this one is about task logging.

Also, usually there is a rule where you don't want double negative in flags which confuses users.
what about "enable_task_logging" with default == True? to make the flag clearer?

gbin (changes requested):

gbin:
@AS1100K Thank you for tackling this one!

gbin:
perfect, I'll pick up the little clippy error in CI so we can merge it, thanks!

gbin (approved):

PR #201, opened 2025-01-05
Host and Device Memory Pool with initial CUDA support + port of existing code.

gbin:
This is a new API to enable heterogeneous memory pool support.


core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
+
+        owned_object.map(|reusable| CuHandle(Arc::new(Mutex::new(CuHandleInner::Pooled(reusable)))))
+    }
+
+    fn space_left(&self) -> usize {
+        self.pool.len()
+    }
+
+    fn copy_from<O: ArrayLike<Element = T::Element>>(&self, from: &mut CuHandle<O>) -> CuHandle<T> {
+        let to_handle = self.acquire().expect("No available buffers in the pool");
```
mik90, line 170: copy_from should probably return a result, i imagine that running out of buffers isn't hard to do. A try and non-try function would make it accessible. The non-try could be implemented as a default on top of the try since it'd just be doing an `expect` on the error

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
+    #[derive(Debug)]
+    pub struct CudaSliceWrapper<E>(CudaSlice<E>);
+
+    impl<E> Deref for CudaSliceWrapper<E>
+    where
+        E: ElementType,
+    {
+        type Target = [E];
+
+        fn deref(&self) -> &Self::Target {
```
mik90, line 227: Haven't really thought this one through, but: It'd be nice to have some way to avoid implementing deref for this. Maybe we can have some other operation that supplies a host memory pool as an arg so it could be implemented for the cuda wrapper. For a host pool, it'd just refer to itself.

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
+                        device
+                            .alloc_zeros(nb_element_per_buffer)
+                            .expect("Failed to allocate device memory"),
+                    )
+                })),
+            }
+        }
+
+        #[allow(dead_code)]
+        pub fn new_with<F>(device: Arc<CudaDevice>, nb_buffers: usize, f: F) -> Self
```
mik90, line 284: Minor: F is the pool initializer callback (https://docs.rs/object-pool/latest/object_pool/struct.Pool.html), it'd be easier to read if the arg names were something like `pool_builder` or something

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
+                .acquire()
+                .expect("No available buffers in the pool");
+
+            match device_handle.lock().unwrap().deref() {
+                CuHandleInner::Pooled(source) => {
+                    match destination_handle.lock().unwrap().deref_mut() {
+                        CuHandleInner::Pooled(ref mut destination) => {
+                            self.device
+                                .dtoh_sync_copy_into(source.as_cuda_slice(), destination)
+                                .expect("Failed to copy data to device");
```
mik90, line 366: We're copying to the host in this function

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
+    where
+        O: ArrayLike<Element = T::Element>;
+}
+
+pub trait DeviceCuPool<T: ArrayLike> {
+    type O: ArrayLike<Element = T::Element>;
+
+    /// Takes a handle to a device buffer and copies it into a host buffer pool.
+    /// It returns a new handle from the host pool with the data from the device handle given.
+    fn copy_into(
```
mik90, line 136: If this was named `copy_to_host` it'd avoid the need to write `/// Copy from device to host` for the implementations on it. Ditto for the opposite route. I'd read "copy_into" as copying into the device pool as opposed to copying into a host pool

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
+    }
+}
 
-        let counters = (0..buffer_count)
-            .map(|_| AtomicUsize::new(0))
-            .collect::<Vec<_>>()
-            .into_boxed_slice();
+#[derive(Debug)]
+/// A buffer that is aligned to a specific size with the Element of type E.
+pub struct AlignedBuffer<E: ElementType> {
```
mik90, line 398: Do we plan on aligning to sizes other than the size of E? In https://users.rust-lang.org/t/how-can-i-allocate-aligned-memory-in-rust/33293/2, it's claimed that Vec already aligns to the size of the instantiated type. The vec can be realigned with `align_to` https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.align_to if the type size isn't known till later

I'm wondering if we can get rid of AlignedBuffer and just use Vec

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
-                drop(handle);
-            } else {
-                handles.push(handle);
-            }
-        }
-        for i in 0..5 {
-            let mut handle = CuHostMemoryPool::allocate(&pool).unwrap();
-            handle.as_slice_mut()[0] = 10 - i;
-            handles.push(handle);
+    #[ignore] // Can only be executed if a real CUDA device is present
```
mik90, line 480: is the `ignore` needed for these tests? I imagine that the `cfg` macro is already limiting it to cuda-only builds

Also, we could just only configure this for cuda-enabled builds and only use cuda on linux. It's certainly possible to use it on windows if anyone feels like fleshing that out

components/sources/cu_v4l/src/lib.rs
```diff
@@ -67,7 +66,7 @@ mod linux_impl {
     }
 
     impl<'cl> CuSrcTask<'cl> for V4l {
-        type Output = output_msg!('cl, CuImage);
+        type Output = output_msg!('cl, CuImage<Vec<u8>>);
```
mik90, line 69: It'd be interesting to have a memory pool as a task itself where requests for memory come in and handles come out. The lack of synchronicity would be a bit awkward but the messaging system would handle concurrent buffer requests coming in at the same time

mik90 (commented):
Leaving some general comments as requested

core/cu29_runtime/src/pool.rs
```diff
@@ -1,235 +1,536 @@
...
         }
     }
 }
 
-pub struct CuBufferHandle {
-    index: usize,
-    pool: Weak<CuHostMemoryPool>,
+/// A shareable handle to an Array coming from a pool (either host or device).
+#[derive(Clone, Debug)]
+pub struct CuHandle<T: ArrayLike>(Arc<Mutex<CuHandleInner<T>>>);
```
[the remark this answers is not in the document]
gbin: yes. Maybe the exception is if you have some driver trying to do some async operation on it. 
so my jam here is: if you have a 4MB matrix or image, a lock of few microsec for a multi ms action on it will be negligible. If you need totally lockfree action then you use the copperlist (but then you might need to copy). There is no free lunch :P

core/cu29_runtime/src/pool.rs
gbin: done

core/cu29_runtime/src/pool.rs
gbin: ha yes, initially I put it in the generic pool but it might move to a "DevicePool" at some point, renaming that for cuda is good enough for now.

core/cu29_runtime/src/pool.rs
gbin: yes, for example if we start to have shared memory buffer, GPU alignments are big.
Also if we want to memory map them, it has to be 16K on ARM with large pages etc...

PR #202, opened 2025-01-07
Add applicative test to hesai driver

luc-k1:


gbin:
The failure with Windows is linked to temp directories: https://github.com/copper-project/copper-rs/actions/runs/12653267696/job/35266963460?pr=202#step:14:470

```rust
    let tmp_dir = tempfile::TempDir::new().expect("could not create a tmp dir");
    let logger_path = tmp_dir.path().join("livoxtester.copper");
```
with

```toml
tempfile = {workspace = true}
```
in cargo.toml

should fix it

Issue #143, opened 2024-12-10
Support mcap export format


gbin:
[mcap](https://mcap.dev/) is a standard logging format used by log visualization and management tools.

We have a placeholder for log extraction here https://github.com/copper-project/copper-rs/blob/master/core/cu29_export/src/lib.rs#L82

We would need to extend that to output a properly formed mcap file.

The rust bindings are here: https://docs.rs/mcap/latest/mcap/

eggyal:
Hi, I'm interested in tackling this one.

Having delved a little into the codebase, it appears that while the parameters of existing log entries are typed (`cu29_value::Value`), the log entries themselves are untyped/schemaless.  (Of course they do have a format into which a string rendering of each parameter is interpolated, but at best that format can only be used to determine how many parameters are expected: not what type they are/should be).

Is this a correct understanding of the status quo?

If so, unless it's possible to construct a schema from some other (out of band) source, it seems to me that:

1. we can map each distinct `CuLogEntry.msg_index` to a distinct `mcap::Channel`;
2. those channels can either be schemaless (not terribly useful) or else we can attempt to infer their schema from the log entry's format string and actual parameter values.  Unfortunately, inference might not always work:
    * suppose a parameter happens to be an *empty* `Value::Option` or `Value::Map`: it will not be possible to ascertain what the type would have been if non-empty—we could continue scanning the log until a non-empty value of that parameter is found but this could be expensive if the logs are large, and it is not guaranteed that any such parameter will ever be found
    * should, for some reason, different entries have different value types for the same parameter, any such inferred schema would be violated: is it possible that this could arise (essentially requiring some sort of union type in the schema) or would this be a fatal error?
3. Alternatively, we could add schema information to Copper's own log format (or use `mcap` internally).

Is this something that the project already has some thoughts about?  My gut says using `mcap` internally might actually be quite sensible: but obviously that'd be a much more significant change that touches on many more areas of the codebase (and requires some consideration regarding migration/ongoing support of existing deployments/log files).

gbin:
Hey @eggyal thank you for checking this thing out!

What about doing something like this: https://github.com/copper-project/copper-rs/pull/151 and asking our users to just add Schema on their messages (maybe optionally or not) so we can export it?

gbin:
using mcap internally btw is not really an option as there are allocations & jitter generation in the lib. It is also 3 to 4x slower than the pretty brutal memmap from copper :P.
[...]

PR #204, opened 2025-01-09
Make ScopedAllocCounter public and add documentation

### LINKED ISSUE #152: Make the ScopedAllocCounter public and document it
As Copper tries to be 0 allocation end to end during the execution time we developed a simple Allocation tracer to monitor that within Copper (and to eventually inform the tasks developers if they do allocate).

But it would be useful externally to, for example for benchmarks, because it was not well exposed for a quick benchmark I had to hack it from the GLOBAL here https://github.com/copper-project/copper-rs/issues/143

Let fix that and make it a first class member of the Copper libraries.

Zac8668:
Makes `ScopedAllocCounter` have some public functions so that it can be used externally.

Closes #152


Zac8668:
Not sure if this is the implementation we want or if it's all of it! Added a example for now that I got from #143 but I can make a more concise one, or maybe just a Doctest would suffice? Any examples to how this stuff is done on Copper are appreciated!

gbin:
yes simply

gbin:
Thanks!

PR #206, opened 2025-01-10
Improve the logging parameters user friendiness

### LINKED ISSUE #205: Improve the logging parameters sanity reporting
Context: Today we have a 2 level logging system that creates "slabs" in one go, map them in memory and do some 0 copy magic into it.

Within this preallocated space we have "sections" which are basically a link list of linear space where various parts of the copper can log into "streams" that create those sections.

for example we have a section type for the main data logger serializing CopperLists (the output of all the tasks) but also we have a section type for the Structured logger (the debug logger).

Those stream get as a parameter how long their sections will be. This is a tuning parameter, the longer they are the riskier it is for your data, the shorter they are the more overhead it will create to constantly create those sections.

Now the issues:
1. If any section preallocation is larger than a Slab size -> Copper will panic
2. If any serialization of a copper list or a structured logging entry is larger that a section -> Copper will panic

So here we would need a proper test set to trigger all those impossible conditions and be sure we report to the user what is going on and some guidance: ie. "The section size of the Data logger is X, you cannot use a slab size smaller than Y, adjust the parameter yaddi yadda"

AS1100K:
Fixes #205 

Also, I Fixed incorrect documentation for `read_configuration_str` as it reads from the String not the file [5c9d82](https://github.com/as1100k-forks/copper-rs/commit/5c9d8242148851d51ea6c2568e7cd15013e91bbb)

### DIFF
--- a/core/cu29_derive/src/lib.rs
+++ b/core/cu29_derive/src/lib.rs
@@ -54,8 +54,9 @@ pub fn gen_cumsgs(config_path_lit: TokenStream) -> TokenStream {
     }
     #[cfg(feature = "macro_debug")]
     eprintln!("[gen culist support with {:?}]", config);
-    let Ok(cuconfig) = read_config(&config) else {
-        return return_error(format!("Failed to read the configuration file: {}", config));
+    let cuconfig = match read_config(&config) {
+        Ok(cuconfig) => cuconfig,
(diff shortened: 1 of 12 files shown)


gbin (approved):

AS1100K:
Yes

PR #212, opened 2025-01-10
Catalog and catch the main proc macro usage error

### LINKED ISSUE #154: Catalog and catch the main proc macro usage error before it blows up in the compilation of the macro
Today if the user makes a mistake in the config for example, it doesn't compile the macro so the error is always on the macro invocation and totally dependent on what the macro is trying to compile from the broken input.

AS1100K:
Fixes #154 

This PR also makes error with RON files much helpful as it not shows, where the error. The following is the error message generated with mistake in RON file

```
error: proc macro panicked
 --> examples/cu_caterpillar/src/logreader.rs:4:1
  |
4 | gen_cumsgs!("copperconfig.ron");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: Syntax Error in config: Expected string at position 4:17
```

### DIFF
--- a/core/cu29_derive/Cargo.toml
+++ b/core/cu29_derive/Cargo.toml
@@ -28,6 +28,10 @@ convert_case = "0.6.0"
 [build-dependencies]
 cu29-unifiedlog = { workspace = true }  # needed
 
+[dev-dependencies]
+trybuild = "1.0"
+cu29 = { workspace = true } # needed for compile_fail tests
+
 [features]
 default = []
 # enables a more verbose build log showing the code generation.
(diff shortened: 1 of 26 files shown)


AS1100K:
I have added catalog of error which came to my mind and added them in `tests/compile_fail` directory.

Also, currently the error in the RON file is shown with line and column position only when it is an syntax error and in order to identify that a certain type exists in the RON file and in the crate would be much difficult but is possible by validating RON file in `build.rs`, atm I am thinking of using rustdoc-json could address this. But I believe this would be too much.

@gbin What are you thoughts on this PR?

gbin:
@AS1100K this is awesome! We will see if the users encounter new cases but this is a really good start.

PR #214, opened 2025-01-11
[FIX] Free space for Ubuntu runner

makeecat:


makeecat:
We will go with splitting the CI into build and release configurtation. Thus we do not need to use the Ubuntu runner cleaner, though this cleaner can free up 17GB with 55s.

gbin:
Thanks for jumping on this one

PR #217, opened 2025-01-12
[FIX] (CI) rust-cache key and fix audit

makeecat:
1. Add matrix.mode in cache key. Default cache key cause conflict when build/release try to save the same cache key.
2. Further simplify the CI, test all features directly.
3. Addressed dependency security issue of cu29_helpers thanks to cargo audit. Use tempfile instead of tempdir. The tempdir was merged into tempfile and became unmaintained.

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -37,7 +37,7 @@ jobs:
         if: runner.os == 'Windows'
         uses: Cyberboss/install-winget@v1
 
-      - name: Cargo fmt check
+      - name: Check formatting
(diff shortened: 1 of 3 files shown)


gbin (approved):

.github/workflows/general.yml
```diff
@@ -166,3 +104,13 @@ jobs:
       - uses: crate-ci/typos@master
         with:
           config: ./.config/_typos.toml
+
+  audit:
```
gbin, line 108: That could be done once a week like we mentioned on discord.

PR #219, opened 2025-01-13
Add debug output pane to cu-consolemon

### LINKED ISSUE #153: Provide a new pane with the debug and stdout/stderr output for cu-consolemon
It would help a lot to be able to use cu-monitor without sacrificing the text output that is logged underneath

AS1100K:
Fixes #153 

## Demo

https://github.com/user-attachments/assets/7ef668f1-3d71-4149-a15f-e3cf1a49aaae

### DIFF
--- a/components/monitors/cu_consolemon/Cargo.toml
+++ b/components/monitors/cu_consolemon/Cargo.toml
@@ -10,8 +10,14 @@ categories.workspace = true
 homepage.workspace = true
 repository.workspace = true
 
+[features]
+default = ["debug_pane"]
+# Enables 'Debug Output' Pane
+debug_pane = ["dep:log", "dep:chrono", "dep:dashmap"]
+
 [dependencies]
 cu29 = { workspace = true }
+cu29-log-runtime = { workspace = true }
 compact_str = { workspace = true }
 ratatui = "0.29"
 pfetch-logo-parser = "0.1"
@@ -23,3 +29,6 @@ tui-nodes = "0.8"
 tui-widgets = "0.4"  # 0.4.0 brings ratatui 0.29
 color-eyre = "0.6"
 gag = "1.0.0"
+log = { version = "0.4", optional = true }
+chrono = { version = "0.4", optional = true }
+dashmap = { version = "6.1", optional = true }
--- a/components/monitors/cu_consolemon/src/lib.rs
+++ b/components/monitors/cu_consolemon/src/lib.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "debug_pane")]
+mod debug_pane;
 pub mod sysinfo;
 
 use ansi_to_tui::IntoText;
@@ -31,16 +33,8 @@ use std::time::Duration;
 use std::{io, thread};
 use tui_nodes::{Connection, NodeGraph, NodeLayout};
 use tui_widgets::scrollview::{ScrollView, ScrollViewState};
-
 #[cfg(feature = "debug_pane")]
-use {
-    compact_str::CompactStringExt,
-    log::{Level, LevelFilter, Log, Metadata, Record},
-    std::collections::VecDeque,
-    std::io::Read,
-    std::sync::atomic::AtomicU16,
-    std::sync::mpsc::{Receiver, SyncSender},
-};
+use {debug_pane::UIExt, log::Log};
 
 #[cfg(feature = "debug_pane")]
 const MENU_CONTENT: &str = "   [1] SysInfo  [2] DAG  [3] Latencies  [4] Debug Output  [q] Quit   ";
@@ -301,100 +295,7 @@ struct UI {
     #[cfg(feature = "debug_pane")]
     error_redirect: gag::BufferRedirect,
     #[cfg(feature = "debug_pane")]
-    debug_output: DebugLog,
-}
-
-#[cfg(feature = "debug_pane")]
-struct DebugLog {
-    debug_log: VecDeque<String>,
-    max_rows: AtomicU16,
-    rx: Receiver<String>,
-}
-
-#[cfg(feature = "debug_pane")]
-impl DebugLog {
-    fn new(max_lines: u16) -> (Self, SyncSender<String>) {
-        let (tx, rx) = std::sync::mpsc::sync_channel(1000);
-        (
-            Self {
-                debug_log: VecDeque::new(),
-                max_rows: AtomicU16::new(max_lines),
-                rx,
-            },
-            tx,
-        )
-    }
-
-    fn push_logs(&mut self, logs: String) {
-        if logs.is_empty() {
-            return;
-        }
-
-        self.debug_log.push_back(logs);
-        let max_row = self.max_rows.load(Ordering::SeqCst) as usize;
-        while self.debug_log.len() > max_row {
-            self.debug_log.pop_front();
-        }
-    }
-
-    fn update_logs(&mut self) {
-        let max_row = self.max_rows.load(Ordering::SeqCst) as usize;
-
-        for log in self.rx.try_iter() {
-            if log.is_empty() {
-                continue;
-            }
-
-            self.debug_log.push_back(log);
-            if self.debug_log.len() > max_row {
-                self.debug_log.pop_front();
-            }
-        }
-    }
-
-    fn get_logs(&mut self) -> String {
-        self.update_logs();
-        self.debug_log.join_compact("").to_string()
-    }
-}
-
-#[cfg(feature = "debug_pane")]
-#[derive(Clone)]
-struct LogSubscriber {
-    tx: SyncSender<String>,
-}
-
-#[cfg(feature = "debug_pane")]
-impl LogSubscriber {
-    fn new(tx: SyncSender<String>) -> Self {
-        let log_subscriber = Self { tx };
-        log::set_boxed_logger(Box::new(log_subscriber.clone())).unwrap();
-        log::set_max_level(LevelFilter::Info);
-        log_subscriber
-    }
-
-    fn push_logs(&self, log: String) {
-        if let Err(err) = self.tx.send(log) {
-            eprintln!("Error Sending Logs to MPSC Channel: {}", err.0)
-        }
-    }
-}
-
-#[cfg(feature = "debug_pane")]
-impl Log for LogSubscriber {
-    fn enabled(&self, metadata: &Metadata) -> bool {
-        metadata.level() <= Level::Debug
-    }
-
-    fn log(&self, record: &Record) {
-        if self.enabled(record.metadata()) {
-            let message = format!("[{}] - {}\n", record.level(), record.args());
-
-            self.push_logs(message);
-        }
-    }
-
-    fn flush(&self) {}
+    debug_output: debug_pane::DebugLog,
 }
 
 impl UI {
@@ -405,7 +306,7 @@ impl UI {
         task_stats: Arc<Mutex<TaskStats>>,
         task_statuses: Arc<Mutex<Vec<TaskStatus>>>,
         error_redirect: gag::BufferRedirect,
-        debug_output: DebugLog,
+        debug_output: debug_pane::DebugLog,
     ) -> UI {
         init_error_hooks();
         let nodes_scrollable_widget_state =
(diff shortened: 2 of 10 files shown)


components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -535,6 +551,20 @@ impl UI {
...
+
+    fn capture_debug_output(&self) {
+        let debug_output = self.debug_output.clone();
+        thread::spawn(move || {
+            let mut buffer = String::new();
+            loop {
+                let mut line = String::new();
+                if let Ok(_) = io::stdin().read_line(&mut line) {
+                    buffer.push_str(&line);
+                    *debug_output.lock().unwrap() = buffer.clone();
```
gbin, line 563: why not `*debug_output.lock().unwrap().push_str(&line)` to avoid a copy?

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -535,6 +551,20 @@ impl UI {
...
+                }
+            }
```
gbin, line 565: then we need something that finds the end of lines and keep a fixed amount of logging in there.
Maybe we need a circular buffer of lines even so we don't copy or move megabytes of text from time to time (adding jitter).

gbin (commented):
Thanks, this is a good start.

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -282,6 +288,147 @@ struct UI {
...
+    }
+
+    fn push_logs<A: AsRef<str>>(&self, logs: A) {
+        let mut log_lines = self.debug_log.lock().unwrap();
+        let max_col = *self.max_col.lock().unwrap();
+        log_lines.push(logs.as_ref().to_string());
+        if log_lines.len() > max_col as usize {
+            log_lines.remove(0);
+        }
+    }
```
makeecat, line 315: Should we use VecDeque instead of Vec for debug_log for performance consideration?

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -282,6 +288,147 @@ struct UI {
...
+
+    fn get_logs(&self) -> std::sync::MutexGuard<'_, Vec<String>> {
+        self.debug_log.lock().unwrap()
+    }
+}
+
+struct LogSubscriber {
+    debug_output: Arc<DebugLog>,
+    span_registry: Arc<Mutex<HashMap<MetadataKey, Id>>>,
```
makeecat, line 324: Should we consider using https://github.com/xacrimon/dashmap for concurrent hash map?

AS1100K:
@makeecat Thanks for the review, I have now used `VecDeque` and `DashMap`. I have also, migrated from using `Mutex` to `Channel`. I did this, as when using `Mutex` there are a lot of instances of long blocking, which makes the application feel laggy. Using Channel blocking has been prevented and thus also improves performance as now the single thread read and writes the log without attaining any lock.

Also, I think we should make this Debug pane behind feature gate as this library is using tracing [`set_global_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html) which is not meant to be used by libraries as if the main crate maybe wants to send these logs or store them somewhere else but then they will not be able to use monitor.

I thinking adding a default feature like `debug-pane` would be good and will introduce flexibility for the main crate user on log behaviour.

makeecat:
I agree with adding a debug-panel faeture pane. In the future we may want to use tokio async to handle all IO-related operations to avoid overhead, but we can start with this implementation at the current stage.

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -282,18 +302,223 @@ struct UI {
...
+    #[cfg(feature = "debug_pane")]
+    error_redirect: gag::BufferRedirect,
+    #[cfg(feature = "debug_pane")]
+    debug_output: DebugLog,
+}
+
+#[cfg(feature = "debug_pane")]
+struct DebugLog {
+    debug_log: VecDeque<String>,
+    max_rows: Arc<Mutex<u16>>,
```
gbin, line 314: can we get away with a simple Mutex as we only protect a u16? Arc is usually when you want to ship the reference around. Even better an AtomicU16 :)

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -282,18 +302,223 @@ struct UI {
...
+        self.update_logs();
+        self.debug_log.join_compact("").to_string()
+    }
+}
+
+#[cfg(feature = "debug_pane")]
+struct LogSubscriber {
+    tx: SyncSender<String>,
+    span_registry: Arc<DashMap<MetadataKey, Id>>,
+    next_id: Arc<Mutex<u64>>,
```
gbin, line 369: same here

components/monitors/cu_consolemon/Cargo.toml
```diff
@@ -23,3 +28,7 @@ tui-nodes = "0.8"
 tui-widgets = "0.4"  # 0.4.0 brings ratatui 0.29
 color-eyre = "0.6"
 gag = "1.0.0"
+tracing = { version = "0.1", optional = true }
```
gbin, line 31: candid question: why do we need tracing? 

This adds another logging facility aside the one from Copper itself [cu29-log](https://www.copper-robotics.com/whats-new/deep-dive-2-structured-text-logging) where we can hook up the output of our structured logging to the standard [log crate](https://github.com/copper-project/copper-rs/blob/master/core/cu29_log_runtime/src/lib.rs#L132) if the user really wants live text logs (only in debug + a flag).

^ why not splitting it in 2, one that subscribed to the log we already use in Copper and an stderr/stdout capture.

Maybe to make it fancy: clearly mark in the ui either with a color or an icon from where the line is coming from (so the user gets a hint if this is a dependency going to std or the standard rust logging)

then the pane is just a list of ANSI Strings to render from the bottom.

gbin: BTW: if we want to also support tracing as an output of the structured logging in Copper, that is another story, I would be supportive as it brings value to users who are already using tracing.

gbin (changes requested):

gbin:
Thanks @AS1100K! My biggest question here is why do we want to introduce tracing in the middle of it. Also, I don't see where we keep the buffer from not growing indefinitely, we should have *somewhere* a constant or parameter saying: we keep the last 1000 lines of back buffer. Otherwise your robot will run out of memory at some point :)

components/monitors/cu_consolemon/Cargo.toml
```diff
@@ -10,6 +10,11 @@ categories.workspace = true
 homepage.workspace = true
 repository.workspace = true
 
+[features]
+default = ["debug_pane"]
```
gbin, line 14: nice call here!

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -570,16 +823,48 @@ impl CuMonitor for CuConsoleMon {
...
+
+                tracing_log::LogTracer::init().expect("Failed to setup log tracer");
+                let log_subscriber = LogSubscriber::new(tx);
+                tracing::subscriber::set_global_default(log_subscriber)
+                    .expect("Failed to set global subscriber");
+
+                ui.run_app(&mut terminal).expect("Failed to run app");
+            }
+
+            #[cfg(not(feature = "debug_pane"))]
```
gbin, line 854:
```suggestion
            #[cfg(not(feature = "debug_pane"))]
            {
```

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -570,16 +823,48 @@ impl CuMonitor for CuConsoleMon {
...
+                let mut ui = UI::new(config_dup, taskids, task_stats_ui, error_states);
+
+                ui.run_app(&mut terminal).expect("Failed to run app");
+            }
+
             quitting.store(true, Ordering::SeqCst);
             // restoring the terminal
-            drop(print_gag);
+            #[cfg(not(feature = "debug_pane"))]
+            drop(stderr_gag);
```
gbin, line 867:
```suggestion
           }
```

components/monitors/cu_consolemon/src/lib.rs
gbin: Just a simple scope would work no?

AS1100K:
> Also, I don't see where we keep the buffer from not growing indefinitely

I also kept that in mind, and only keep the number of logs according to terminal height. here https://github.com/copper-project/copper-rs/pull/219/files#diff-77cbd66ea06743fe2a1e13a87d5525ab87330d6cc6d548f35068b256168b611cR352-R356

gbin:
> > Also, I don't see where we keep the buffer from not growing indefinitely
> 
> I also kept that in mind, and only keep the number of logs according to terminal height. here https://github.com/copper-project/copper-rs/pull/219/files#diff-77cbd66ea06743fe2a1e13a87d5525ab87330d6cc6d548f35068b256168b611cR352-R356

aaaah... ok ... we need a back buffer :) people will need to scroll back up a little bit

components/monitors/cu_consolemon/Cargo.toml
AS1100K: > stdout capture

The problem is we can't listen to `stdout` as locking `stdin` would make the entire terminal irresponsive and will also capture `stdout` send by `crossterm` which will mess the entire stuff. Therefore, I went by adding support for `tracing` and `log` _(via `tracing-log`)_ crate as both `tracing` and `log` are very popular logging frameworks and this will allow any other libraries logs to be seen by the end user.

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -282,18 +302,223 @@ struct UI {
...
+    fn generate_id(&self) -> Id {
+        let mut next_id = self.next_id.lock().unwrap();
+        let id = *next_id;
+        *next_id += 1;
+        Id::from_u64(id)
+    }
+}
+
+#[cfg(feature = "debug_pane")]
+impl Subscriber for LogSubscriber {
```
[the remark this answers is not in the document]
AS1100K: > This adds a lot of complexity

For most of the stuff we could just have `event` function and have other functions `unimplemented!()` but this would become a issue if some libraries are using `span` macro as the other functions in the trait are just to support [span](https://docs.rs/tracing/latest/tracing/span/index.html) events

components/monitors/cu_consolemon/Cargo.toml
gbin: yeah but we don't want tracing on the main stack, remember Copper is 0 allocation after the tasks area starting, we don't even construct Strings. This is why we implemented the structured logging.

AS1100K:
@gbin Thanks for you review, I have now removed the support for `tracing` crate and now it supports `cu29-log` and `log` crate. Also, while doing so I introduced a new argument in [`basic_copper_setup`](https://github.com/copper-project/copper-rs/compare/master...AS1100K:copper-rs:add-debug-output#diff-bc7e34b15032b0327094bade29a60c828a1256f7bc561297a6046acf3e8163d9R31) and [`LoggerRuntime::init`](https://github.com/copper-project/copper-rs/compare/master...AS1100K:copper-rs:add-debug-output#diff-6356a806527043ab67b7bf477ac642af378e7e4f1a632be69e07b54729239e23R61) function. I would love some feedback here as I am not sure if this is a right call here.



> > > Also, I don't see where we keep the buffer from not growing indefinitely
> > 
> > 
> > I also kept that in mind, and only keep the number of logs according to terminal height. here https://github.com/copper-project/copper-rs/pull/219/files#diff-77cbd66ea06743fe2a1e13a87d5525ab87330d6cc6d548f35068b256168b611cR352-R356
> 
> aaaah... ok ... we need a back buffer :) people will need to scroll back up a little bit

Also, I would also implement it soon

gbin:
Cool! Thank you for your patience!

cu-consolemon is not part of core, I don't think it should: for example in production robots should not have it and might use a different implementation behind the Monitoring API.

what about hooking that up either as a Log destination (that will be picked up when consolemon is in) or, another possible way is to add a logging API to the monitoring API (that consolemon relies on and decouples it from core)?

components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -282,18 +298,141 @@ struct UI {
     sysinfo: String,
     task_stats: Arc<Mutex<TaskStats>>,
     nodes_scrollable_widget_state: NodesScrollableWidgetState,
+    #[cfg(feature = "debug_pane")]
+    error_redirect: gag::BufferRedirect,
+    #[cfg(feature = "debug_pane")]
+    debug_output: DebugLog,
+}
+
+#[cfg(feature = "debug_pane")]
```
gbin, line 307: put all that in a module and make a use of the module when the feature is on

core/cu29_log_runtime/Cargo.toml
```diff
@@ -21,6 +21,7 @@ cu29-clock = { workspace = true }
 bincode = { workspace = true }
 smallvec = "1.13.2"
 log = "0.4.22"
+once_cell = "1.20"
```
gbin, line 24: this is not integrated to the stdlib in the recent Rust versions?

gbin (approved):
Ok, 2 comments left and we merge this.

gbin:
Thanks @AS1100K good work!

PR #227, opened 2025-01-18
[FIX] pool cuda support for windows platform

makeecat:
## Issue
1. pool.rs cuda implementation does not build for windows platform in CI (https://github.com/copper-project/copper-rs/actions/runs/12837161641/job/35800273948)
2. the CI does not reflect non-cuda platform setup for linux/windows platform
## Solution
1. fixed cfg setup in pool.rs
2. separate CI for non-CUDA and CUDA platforms
3. set up features for non-CUDA platforms in CI manually to exclude CUDA feature. In the future, we need to check if there are new features being introduced, which we need to add these feature flags into CI for non-CUDA platform.

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -36,22 +36,6 @@ jobs:
       - name: Check formatting
         run: cargo +stable fmt --all -- --check
 
-      - name: Free Disk Space (Ubuntu)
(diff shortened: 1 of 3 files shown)


gbin:
oh ok, that works for me, thanks @makeecat !

PR #228, opened 2025-01-19
Make cu-consolemon debug pane only active if `text_log` is true

AS1100K:
This PR makes the following behaviour work:

```txt
text_log = True with no cu-consolemon -> dumping in the console like it was doing before
text_log = False -> ONLY structured logging in the main log no text is constructed ever.
text_log = True with cu-consolemon  -> redirect that to the Pane
text_log = False with cu-consolemon  -> by order of preference: message in the Pane, no pane or empty Pane it there is no way to know
compilation in release mode -> text_log = False is forced, I would even compile out the support like I did for the standard console version
```
_As discussed on discord_

as well as also add time to the log shown in cu-consolemon

### DIFF
--- a/components/monitors/cu_consolemon/src/debug_pane.rs
+++ b/components/monitors/cu_consolemon/src/debug_pane.rs
@@ -20,6 +20,7 @@ pub struct DebugLog {
 }
 
 impl DebugLog {
+    #[allow(dead_code)]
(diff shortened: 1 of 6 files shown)


gbin:
Thanks

gbin:
image
Minor bug: it looks like it doesn't fully clean the screen (you see a some characters remaining from the neofetch)

PR #242, opened 2025-02-23
Avoids .0 notation

gbin:
This is for all the code except generated code and test code.

### DIFF
--- a/core/cu29_runtime/src/monitoring.rs
+++ b/core/cu29_runtime/src/monitoring.rs
@@ -263,9 +263,10 @@ pub struct CuDurationStatistics {
 
 impl CuDurationStatistics {
     pub fn new(max: CuDuration) -> Self {
+        let CuDuration(max) = max;
         CuDurationStatistics {
-            bare: LiveStatistics::new_with_max(max.0),
-            jitter: LiveStatistics::new_with_max(max.0),
+            bare: LiveStatistics::new_with_max(max),
+            jitter: LiveStatistics::new_with_max(max),
             last_value: CuDuration::default(),
         }
     }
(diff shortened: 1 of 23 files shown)


gbin:
I came across a comment saying that basically you never need the .0 notation in Rust because you can deconstruct the wrappers, as we use wrappers all over the place in Copper I think it makes the code way more readable by giving a name of the wrappee.

examples/cu_pointclouds/Cargo.toml
```diff
@@ -16,4 +16,6 @@ cu29-helpers = { workspace = true }
 tempfile = "3.14.0"
 cu-hesai = { path = "../../components/sources/cu_hesai", version = "0.6.0" }
 cu-udp-inject = { path = "../../components/testing/cu_udp_inject" }
+cu-sensor-payloads = { path = "../../components/payloads/cu_sensor_payloads" }
+uom = { workspace = true }
```
makeecat, line 20: do we need it?

gbin: yeah because some unpacks make the type explicit do you need to be able to use them.

core/cu29_runtime/src/monitoring.rs
```diff
@@ -334,11 +334,12 @@ impl CuDurationStatistics {
     #[inline]
     pub fn record(&mut self, value: CuDuration) {
         let CuDuration(nanos) = value;
-        self.bare.stats.record(nanos).unwrap();
         if self.bare.is_empty() {
+            self.bare.stats.record(nanos).unwrap();
             self.last_value = value;
             return;
         }
+        self.bare.stats.record(nanos).unwrap();
```
makeecat, line 342: Can you add a comment to explain the reason behind this commit? I believe it is not very trivial for future users.

makeecat (changes requested):

core/cu29_runtime/src/monitoring.rs
gbin: Ha yeah it tripped me I introduced a regression because once you record it is not empty anymore :)

gbin: done

PR #251, opened 2025-03-05
Input order of a task is now sorted according to the order of definition of the connections

### LINKED ISSUE #187: The order of the tuples on a task getting inputs from other tasks depends on Copper's internal scheduling
For example: 
`type Input = input_msg!('cl, ImageRGBU8Msg, ImageRGBU8Msg, ImageGrayU8Msg);` or 
`type Input = input_msg!('cl, ImageRGBU8Msg, ImageGrayU8Msg, ImageRGBU8Msg);` for rerun is not obvious at first glance from: 

```RON
(
    tasks: [
        (
            id: "cam",
            type: "copper::tasks::webcam::Webcam",
            config: {
                "camera_id": 0,
                "res_rows": 480,
                "res_cols": 640,
                "fps": 30,
            },
        ),
        (
            id: "cam_garden",
            type: "copper::tasks::rtsp::RtspCamera",
            config: {
                "url": "rtsp://tapo_entrance:123456789@192.168.1.141:554/stream2",
            },
        ),
        (
            id: "rerun",
            type: "copper::tasks::rerun_viz::RerunViz",
        ),
        (
            id: "sobel",
            type: "copper::tasks::sobel::Sobel",
        ),
    ],
    cnx: [
        (src: "cam_garden", dst: "rerun", msg: "copper::tasks::ImageRGBU8Msg"),
        (src: "cam", dst: "rerun", msg: "copper::tasks::ImageRGBU8Msg"),
        (src: "sobel", dst: "rerun", msg: "copper::tasks::ImageGrayU8Msg"),
        (src: "cam", dst: "sobel", msg: "copper::tasks::ImageRGBU8Msg"),
    ],
    logging: (
        slab_size_mib: 1024, // Preallocates 1GiB of memory map file at a time
        section_size_mib: 100, // Preallocates 100MiB of memory map per section for the main logger.
    ),
)
```

hscoelho:
fixes #187 

Took me a while, but finally got some time to work on this 😅. 

I'm pretty confident this solution doesn´t affect other parts of the runtime unexpectedly, since it only mutates the input_msg_indices_types object, but I have made some assumptions to actually obtain the id used to sort and I decided panicking would be more appropriate than having a fallback option, so I'm a little worried there's an edge case I'm not aware that might break this solution. (I did test with the examples though).

The reason I created two unit tests was to make sure this works both with a config file and programmatically.

### DIFF
--- a/core/cu29_runtime/src/config.rs
+++ b/core/cu29_runtime/src/config.rs
@@ -786,4 +786,30 @@ mod tests {
         let config = CuConfig::deserialize_ron(txt);
         assert!(config.validate_logging_config().is_err());
     }
+
+    // this test makes sure the edge id is suitable to be used to sort the inputs of a task
+    #[test]
+    fn test_deserialization_edge_id_assignment() {
+        // note here that the src1 task is added before src2 in the tasks array,
+        // however, src1 connection is added AFTER src2 in the cnx array
+        let txt = r#"( 
+            tasks: [(id: "src1", type: "a"), (id: "src2", type: "b"), (id: "sink", type: "c")],
+            cnx: [(src: "src2", dst: "sink", msg: "msg1"), (src: "src1", dst: "sink", msg: "msg2")]
+        )"#;
+        let config = CuConfig::deserialize_ron(txt);
+        assert!(config.validate_logging_config().is_ok());
+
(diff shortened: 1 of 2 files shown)


gbin:
fantastic! I'll take a look tomorrow morning

PR #266, opened 2025-03-25
Added a design doc for the mission

gbin:


gbin:
This should address the comments I had from various parties

PR #283, opened 2025-04-14
[CHORE] update feature deps with 'dep:' and remove unused ones

makeecat:


makeecat:
@gbin Could you check if it is necessary to include gstreamer feature and dependency in cu29_runtime and cu_sensor_payload?

gbin:
> @gbin Could you check if it is necessary to include gstreamer feature and dependency in cu29_runtime and cu_sensor_payload?

@makeecat A little bit like the cuda stuff, if someone doesn't install gstreamer and I don't want Copper compilation failing (in general I want to be sure we are as isolated from system dependencies as possible, learning from ROS :P ). Maybe we could structure that as a simple dependency instead?

makeecat:
> > @gbin Could you check if it is necessary to include gstreamer feature and dependency in cu29_runtime and cu_sensor_payload?
> 
> @makeecat A little bit like the cuda stuff, if someone doesn't install gstreamer and I don't want Copper compilation failing (in general I want to be sure we are as isolated from system dependencies as possible, learning from ROS :P ). Maybe we could structure that as a simple dependency instead?

I understood the intention. How about:
1. add a workspace feature flag gst: if gst is not enabled, we should ignore `cu_gstreamer` package.
2. remove the feature flag gst and dependency of gstreamer and gstreamer_app from `cu29_runtime` and `cu_sensor_payload`

I can propose a separate PR to fix this.

If you think the rest of the current PR looks good to you, you can merge this PR.

gbin:
ok, I agree.

PR #316, opened 2025-05-06
[FEAT] Add ROS compatible zenoh sink

kamibo:
This PR adds a ROS compatible Zenoh sink to forward data from Copper applications to ROS 2 via the Zenoh middleware (rmw_zenoh).

- Introduces a new sink (cu-zenoh-ros-sink) with full implementation including liveliness, key expression formatting, and configuration handling.
- Adds an example application (cu_zenoh_ros) and updates workspace members and dependencies accordingly.

### DIFF
--- a/components/payloads/cu_ros_payloads/Cargo.toml
+++ b/components/payloads/cu_ros_payloads/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "cu-ros-payloads"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description = "Sepcial copper payloads designed to be ROS compatible"
+
+[dependencies]
+compact_str = { workspace = true }
+serde = { workspace = true }
+cu29-value = { workspace = true }
+cu-sensor-payloads = { workspace = true }
--- a/components/payloads/cu_ros_payloads/src/builtin.rs
+++ b/components/payloads/cu_ros_payloads/src/builtin.rs
@@ -1,0 +1,14 @@
+use compact_str::CompactString;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Header {
+    pub stamp: Time,
+    pub frame_id: CompactString,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Time {
+    pub sec: i32,
+    pub nanosec: u32,
+}
--- a/components/payloads/cu_ros_payloads/src/lib.rs
+++ b/components/payloads/cu_ros_payloads/src/lib.rs
@@ -1,0 +1,30 @@
+pub mod builtin;
+pub mod sensor_msgs;
+pub mod std_msgs;
+
+use serde::Serialize;
+use std::convert::From;
+
+// By default use Rust type as ROS type
+#[macro_export]
+macro_rules! ros_type_name {
+    ($t:ty) => {{
+        std::any::type_name::<$t>().rsplit("::").next().unwrap()
+    }};
+}
+
+/// ROS adaptation trait to convert payload data to ROS compatible message.
+/// The output type must match the structure of the related "msg" file.
+/// The namespace relates to the ROS namespace (such as "std_msgs")
+/// and the type name is the same as the message filename.
+pub trait RosMsgAdapter<'a>: std::marker::Sized {
+    type Output: Serialize + for<'b> From<&'b Self>;
+
+    fn namespace() -> &'a str;
+    fn type_name() -> &'a str {
+        ros_type_name!(Self::Output)
+    }
+    fn convert(&self) -> Self::Output {
+        self.into()
+    }
+}
--- a/components/payloads/cu_ros_payloads/src/sensor_msgs.rs
+++ b/components/payloads/cu_ros_payloads/src/sensor_msgs.rs
@@ -1,0 +1,57 @@
+use compact_str::CompactString;
+use cu_sensor_payloads::PointCloudSoa;
+use serde::{Deserialize, Serialize};
+
+use crate::{builtin::Header, RosMsgAdapter};
+
+// sensor_msgs/PointField
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PointField {
+    pub name: CompactString,
+    pub offset: u32,
+    pub datatype: u8,
+    pub count: u32,
+}
+
+// sensor_msgs/PointCloud2 like struct
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PointCloud2 {
+    pub header: Header,
+    pub height: u32, //
+    pub width: u32,  // Length of the point cloud
+    pub pointfields: Vec<PointField>,
+    pub is_bigendian: bool,
+    pub point_step: u32, // Length of a point in bytes
+    pub row_step: u32,   // Length of a row in bytes
+    pub data: Vec<u8>,   // size is (row_step*height)
+    pub is_dense: bool,  // True if there are no invalid points
+}
+
+impl<const N: usize> RosMsgAdapter<'static> for PointCloudSoa<N> {
+    type Output = PointCloud2;
+
+    fn namespace() -> &'static str {
+        "sensor_msgs"
+    }
+
+    fn type_name() -> &'static str {
+        "PointCloud2"
+    }
+}
+
+impl<const N: usize> From<&PointCloudSoa<N>> for PointCloud2 {
+    #[allow(unreachable_code)]
+    fn from(_: &PointCloudSoa<N>) -> Self {
+        PointCloud2 {
+            header: todo!(),
+            height: 1,
+            width: todo!(),
+            pointfields: todo!(),
+            is_bigendian: true,
+            point_step: todo!(),
+            row_step: todo!(),
+            data: todo!(),
+            is_dense: todo!(),
+        }
+    }
+}
--- a/components/payloads/cu_ros_payloads/src/std_msgs.rs
+++ b/components/payloads/cu_ros_payloads/src/std_msgs.rs
@@ -1,0 +1,37 @@
+use serde::{Deserialize, Serialize};
+
+use crate::RosMsgAdapter;
+
+// std_msgs/Int8 like struct
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Int8 {
+    pub data: i8,
+}
+
+impl RosMsgAdapter<'static> for i8 {
+    type Output = Int8;
+
+    fn namespace() -> &'static str {
+        "std_msgs"
+    }
+}
+
+impl From<&i8> for Int8 {
+    fn from(value: &i8) -> Self {
+        Self { data: *value }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_adapter_type_info() {
+        let x: i8 = 42;
+        let msg = x.convert();
+        assert_eq!(msg.data, 42);
+        assert_eq!(<i8 as RosMsgAdapter>::namespace(), "std_msgs");
+        assert_eq!(<i8 as RosMsgAdapter>::type_name(), "Int8");
+    }
+}
--- a/components/sinks/cu_zenoh_ros_sink/README.md
+++ b/components/sinks/cu_zenoh_ros_sink/README.md
@@ -1,0 +1,65 @@
+# cu-zenoh-ros-sink
+
+This is a Copper sink task that publishes messages to ROS 2 via Zenoh middleware.
+It allows seamless integration between Copper applications and ROS 2 nodes by using the [RMW zenoh implementation](https://github.com/ros2/rmw_zenoh).
+
+## Configuration
+
+The sink requires the following configuration parameters:
+
+- zenoh_config_json: Zenoh config [as json string](https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5) (optional).
+- domain_id: ROS domain ID (optional, defaults to "0"),
+- namespace: The ROS namespace (optional, defaults to "copper").
+- node: The ROS node name (optional, defaults to "node").
+- topic: The ROS topic name to publish to (optional, defaults to "copper").
+
+Example configuration in a Copper application:
+
+```ron
+(
+    tasks: [
+        (
+            id: "publisher",
+            config: {
+                "domain_id": 0,
+                "namespace": "copper",
+                "node": "sink_node",
+                "topic": "output",
+            }
+        ),
+    ],
+    // ... rest of the configuration
+)
+```
+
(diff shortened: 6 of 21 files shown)


makeecat (changes requested):
This PR adds a new zenoh sink component that allows Copper to publish data to ROS 2 using the Zenoh middleware. This enables seamless integration between Copper applications and ROS 2 nodes, using the RMW zenoh implementation.

Several Actions items:
1. Consistency of Error Msg: Please either use `ZenohRosSink` or `ZenohSink` as the prefix of error msg.
2. Add more documentation for public API
3. Fix `zenoh` version back to `1.3.4` to avoid surprise
4. The CDR serialization adds an additional layer of conversion that could impact performance, worth noting it in README for future improvements.
5. In `components/sinks/cu_zenoh_ros_sink/src/lib.rs` (process method), CDR serialization uses `.unwrap()` which can panic, please handle the error:
  `let serial_ros_payload = cdr::serialize::<_, _, CdrBe>(&ros_payload, Infinite).unwrap();`

kamibo:
> This PR adds a new zenoh sink component that allows Copper to publish data to ROS 2 using the Zenoh middleware. This enables seamless integration between Copper applications and ROS 2 nodes, using the RMW zenoh implementation.
> 
> Several Actions items:
> 
>     1. Consistency of Error Msg: Please either use `ZenohRosSink` or `ZenohSink` as the prefix of error msg.
> 
>     2. Add more documentation for public API
> 
>     3. Fix `zenoh` version back to `1.3.4` to avoid surprise
> 
>     4. The CDR serialization adds an additional layer of conversion that could impact performance, worth noting it in README for future improvements.
> 
>     5. In `components/sinks/cu_zenoh_ros_sink/src/lib.rs` (process method), CDR serialization uses `.unwrap()` which can panic, please handle the error:
>        `let serial_ros_payload = cdr::serialize::<_, _, CdrBe>(&ros_payload, Infinite).unwrap();`

All done.
Regarding the API documentation, I refactored the ROS message adapter trait to make it easier to use and added some comments.
Let me know if there’s anything else we can do to improve it.

makeecat (approved):
Works for me. @gbin Do you have some comments on the use of string ? I believe we can improve that in a second PR to have a baseline.

kamibo:
TIL: They already planned to change the attachment format for the next release https://github.com/ros2/rmw_zenoh/issues/600  :disappointed:

components/payloads/cu_ros_payloads/src/builtin.rs
```diff
@@ -0,0 +1,13 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Header {
+    pub stamp: Time,
+    pub frame_id: std::string::String,
```
gbin, line 6: Nit: if the string is short maybe use a short string like compact_str we use in cu29_runtime?

components/payloads/cu_ros_payloads/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+/// The output type must match the structure of the related "msg" file.
+/// The namespace relates to the ROS namespace (such as "std_msgs")
+/// and the type name is the same as the message filename.
+pub trait RosMsgAdapter {
+    type Output: Serialize;
+
+    fn namespace() -> &'static str;
+    fn type_name() -> &'static str;
+    fn convert(&self) -> Self::Output;
+}
```
gbin, line 17: I am not sure about forcing static here, if you parametrize the type I think the implementor/users of that trait can always use static for it:

```Rust
pub trait RosMsgAdapter<'a> {
    type Output: Serialize;

    fn namespace() -> &'a str;
    fn type_name() -> &'a str;
    fn convert(&self) -> Self::Output;
}
``` 

```Rust
impl<'a> RosMsgAdapter<'a> for MyMsg {
    type Output = MyRosCompatibleMsg;

    fn namespace() -> &'a str {
        "std_msgs"
    }

    fn type_name() -> &'a str {
        "String"
    }

    fn convert(&self) -> Self::Output {
        MyRosCompatibleMsg { ... }
    }
}
```

components/payloads/cu_ros_payloads/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "cu-ros-payloads"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description = "Those are ROS compatible payloads"
```
gbin, line 11: This will be the basic public description on Crates.io, maybe mention that it is part of copper/extra payload for copper.

components/payloads/cu_ros_payloads/src/sensor_msgs.rs
```diff
@@ -0,0 +1,54 @@
+use cu_sensor_payloads::PointCloudSoa;
+use serde::{Deserialize, Serialize};
+
+use crate::{builtin::Header, RosMsgAdapter};
+
+// sensor_msgs/PointField
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PointField {
+    pub name: String,
```
gbin, line 9: maybe small_str

components/payloads/cu_ros_payloads/src/sensor_msgs.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn namespace() -> &'static str {
+        "sensor_msgs"
+    }
+
+    fn type_name() -> &'static str {
+        "PointCloud2"
+    }
+
+    #[allow(unreachable_code)]
+    fn convert(&self) -> Self::Output {
```
gbin, line 41: can't we just use the standard "Into" facility of Rust here?

components/payloads/cu_ros_payloads/src/std_msgs.rs
```diff
@@ -0,0 +1,25 @@
+use serde::{Deserialize, Serialize};
+
+use crate::RosMsgAdapter;
+
+// std_msgs/Int8 like struct
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Int8 {
+    pub data: i8,
+}
```
gbin, line 9: maybe Int8(i8)?

components/payloads/cu_ros_payloads/src/std_msgs.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    fn convert(&self) -> Self::Output {
+        Int8 { data: *self }
+    }
+
+    fn namespace() -> &'static str {
+        "std_msgs"
+    }
+
+    fn type_name() -> &'static str {
```
gbin, line 22: if the type_name is almost always the Struct name you can provide it for the user with a small default impl in the trait maybe? 

```Rust
pub trait RosMsgAdapter<'a> {
    type Output: serde::Serialize;

    fn convert(&self) -> Self::Output;

    fn namespace() -> &'a str;

    fn type_name() -> &'a str {
        ros_type_name!(Self::Output)
    }
}

#[macro_export]
macro_rules! ros_type_name {
    ($t:ty) => {{
        std::any::type_name::<$t>()
            .rsplit("::")
            .next()
            .unwrap()
    }};
}

```

link to the playground with a little test:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c10055c62bf8aadd4e77e9920d3965a7

note that the user can override this behavior in the implementation

components/sinks/cu_zenoh_ros_sink/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Example configuration in a Copper application:
+
+```ron
+(
+    tasks: [
+        (
+            id: "publisher",
+            config: {
+                "zenoh_config_file": "/opt/ros/humble/share/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5",
```
gbin, line 24: That breaks the "self-contained" nature of the Copper config.

We do an include! of the RON file at compile time to be able to have a 1 file deployment and a path to a no_std deployment too: ie. *no* filesystem at all.

is it possible, maybe in another PR to provide an embedded version of it: for example a config_str: and the inline version of the json file?

components/sinks/cu_zenoh_ros_sink/src/error.rs
```diff
@@ -0,0 +1,10 @@
+use cu29::CuError;
+use zenoh::Error as ZenohError;
+
+pub fn cu_error(msg: &str, error: ZenohError) -> CuError {
```
gbin, line 4: Ok those ones are getting ridiculous, I am adding a note to refactor CuError with something like:
```Rust
#[derive(Debug)]
pub struct CuError {
    message: String,
    cause: Option<Box<dyn std::error::Error + Send + Sync>>,
}
```
No action needed on your part yet :)
https://github.com/copper-project/copper-rs/issues/317

components/sinks/cu_zenoh_ros_sink/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    session: zenoh::Config,
+    domain_id: u32,
+    namespace: String,
+    node: String,
+    topic: String,
+}
+
+pub struct ZenohRosContext {
+    session: zenoh::Session,
+    publisher: zenoh::pubsub::Publisher<'static>,
```
gbin, line 50: The publisher has to be static here?

components/sinks/cu_zenoh_ros_sink/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    #[allow(dead_code)]
+    publisher_token: zenoh::liveliness::LivelinessToken,
+    sequence_number: u64,
+}
+
+impl<P> Freezable for ZenohRosSink<P> where P: CuMsgPayload + RosMsgAdapter {}
+
+impl<'cl, P> CuSinkTask<'cl> for ZenohRosSink<P>
+where
+    P: CuMsgPayload + RosMsgAdapter + 'cl + 'static,
```
gbin, line 63: while correct rust, this 'static feels out of place. Conceptually those types only need to live for the duration of one CopperList ie. one execution cycle.

gbin (changes requested):
Awesome PR,

tried to prod on some things what looked weird, feel free to push back.

Thank you so much!

kamibo:
Lots of good points :+1:
Note: I'm going be busy for a week (team meeting in Paris).  I'll continue when I get back

components/payloads/cu_ros_payloads/src/std_msgs.rs
kamibo: It should work but I think it is better to stick to ROS implementation https://docs.ros.org/en/melodic/api/std_msgs/html/msg/Int8.html

components/sinks/cu_zenoh_ros_sink/src/lib.rs
kamibo: Actually, as far as I understand, it only means that the "key expr" is owned. But maybe I'm missing something.

components/payloads/cu_ros_payloads/src/lib.rs
kamibo: Let me known if it sounds better now.

components/sinks/cu_zenoh_ros_sink/src/lib.rs
gbin: So what I understand is that basically either it is a static string OR you won't be able to deallocate the publisher (Box::leak), so the ZenohRosContext, so the session etc.. ?

I believe you need to forward those lifetimes to the Context.

'static is really what it means, it will be never ever be deallocated.

components/payloads/cu_ros_payloads/src/lib.rs
gbin: I was thinking about the convert method ... this is just a not very idiomatic way of a From/Into, you can force the trait with a + 

```
pub trait RosMsgAdapter<'a>: Into<Self::Output> + Sized
where
    Self::Output: Serialize + 'a
{
    fn namespace() -> &'a str;
    fn type_name() -> &'a str {
        ros_type_name!(Self::Output)
    }
}
```
^ that feels more like the signature I would expect

gbin (changes requested):

components/sinks/cu_zenoh_ros_sink/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+                namespace: config.get::<String>("namespace").unwrap_or("node".into()),
+                node: config.get::<String>("node").unwrap_or("node".into()),
+                topic: config.get::<String>("topic").unwrap_or("copper".into()),
+            },
+            ctx: None,
+        })
+    }
+
+    fn start(&mut self, _clock: &RobotClock) -> CuResult<()> {
+        let session = zenoh::Wait::wait(zenoh::open(self.config.session.clone()))
```
edgarriba, line 97: How heavy it’s to create a new session per task, instead of having somehow a global ?

components/sinks/cu_zenoh_ros_sink/src/lib.rs
kamibo: > I believe you need to forward those lifetimes to the Context.

I also tried this but I can't figure out to forward those lifetimes beyond code generated from the RON file.
But I'm not sure it is a big deal.
This lifetime is related to the keyexpr stored inside the publisher and according the documentation
https://docs.rs/zenoh-keyexpr/latest/zenoh_keyexpr/

> [KeyExpr](https://docs.rs/zenoh/latest/zenoh/key_expr/struct.KeyExpr.html) works like a [Cow<str>](https://doc.rust-lang.org/nightly/alloc/borrow/enum.Cow.html), but also stores some additional context internal to  Zenoh to optimize routing and network usage.

AFIU this means that using static lifetime implies key expr cannot be borrowed https://github.com/eclipse-zenoh/zenoh/blob/63a6e33fe58eb98485c13dad0924f673baedc060/zenoh/src/api/key_expr.rs#L40
And we don't care about that as we have to compute the key expr.

> 'static is really what it means, it will be never ever be deallocated.

Not necessarily. My understanding is it can be hold for as long as needed.

https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html#trait-bound

But maybe I'm missing something...

components/sinks/cu_zenoh_ros_sink/src/lib.rs
edgarriba: in a separated projet using copper I have instantiated within a lazy operator

kamibo: This sounds great. However I'm wondering how to define a single zenoh session config for all the tasks inside the RON file? :thinking: 
@gbin any thoughts about this?

gbin: I think we really need a new kind of "service" task just holding a state / running a daemon thread. I propose to add that as an issue for that and discuss the arch and determinism on the issue. For now let's move forward with this PR with the known limitation then refactor it with the new facility in the framework.

edgarriba: really good idea. Looking forward to it

gbin:
Ok let's merge this and go back to it for improvements, I don't want the PR to go stale.

Thanks everyone

PR #324, opened 2025-05-13
[CHORE] replace lazy_static by std::sync::OnceLock

howjmay:
This PR replaces lazy_static! with std::sync::OnceLock, related crates are `cu-rp-gpio` and `cu29-log-derive`.

### DIFF
--- a/components/sinks/cu_rp_gpio/Cargo.toml
+++ b/components/sinks/cu_rp_gpio/Cargo.toml
@@ -19,7 +19,6 @@ cu29-log = { workspace = true }
 cu29-log-runtime = { workspace = true } # needed
 bincode = { workspace = true }
 serde = { workspace = true }
-lazy_static = "1.5.0"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 rppal = { version = "0.22.1", features = ["hal"] }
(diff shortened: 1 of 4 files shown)


gbin:
I wanted to remove this external dependency for a while as it is supported on std. Today we have a mix bag of both so let's go std.

As for no-std, let's split the concerns with the feature flag because there will be other no-std friendly crates we will need to include for stuff like Vec etc ...

So I'd say let's merge this, remove the dependency then maybe we will have to readd it for their spin lock implementation or go another route when we start the no-std work.

gbin:
Thx @howjmay

howjmay:
@makeecat @gbin thank you for merging. 
One question, I want to contribute to this project. As a beginner of the project, may you give me some advice on how I can start. Thanks

makeecat:
> @makeecat @gbin thank you for merging. One question, I want to contribute to this project. As a beginner of the project, may you give me some advice on how I can start. Thanks

1. You can join discord dev channel to propose idea and ask questions.
2. You can look into examples to get more familiar with the framework, cu_rp_balancebot and cu_caterpillar are good ones to start.
3. You can also generate a template project by following instructions in the templates folder.

PR #330, opened 2025-05-15
[FEAT] cu_transform

makeecat:


components/payloads/cu_spatial_payloads/src/lib.rs
edgarriba, line 1: @makeecat @gbin have you considered using something light like `glam-rs` which as all the abstractions with optimizations ?

makeecat: That is a good idea. I am trying to start from a native structure and explore the architecture, in order to understand the potential issues.

This draft will mostly likely be refactored and build from scratch again. Could you also share more alternatives to `glam-rs`?

edgarriba: another alternative is `nalgebra` but that's a larger lib. In our company we ended up defaulting to glam-rs as it simple enough to implement the basics for robotics ops and quite efficiently

components/cu_transform/examples/basic_usage.rs
```diff
@@ -0,0 +1,53 @@
...
+    let mut tree = TransformTree::<f32>::new();
+
+    let world_to_robot = StampedTransform {
+        transform: {
+            let mut t = Transform3D::default();
+            t.mat[0][3] = 1.0; // X translation
+            t.mat[1][3] = 2.0; // Y translation
+            t
+        },
+        stamp: CuDuration(1000),
```
gbin, line 15: would a StampedTransform be simply a CuMsg<Transform>? (with have all the timestamps for the TOV in there or this is another semantic?)

components/cu_transform/examples/basic_usage.rs
```diff
@@ -0,0 +1,53 @@
...
+    let robot_to_camera = StampedTransform {
+        transform: {
+            let mut t = Transform3D::default();
+            t.mat[0][3] = 0.2; // X translation
+            t.mat[1][3] = 0.0; // Y translation
+            t.mat[2][3] = 0.3; // Z translation
+            t
+        },
+        stamp: CuDuration(1000),
+        parent_frame: "robot".to_string(),
```
gbin, line 29: could we use anything that is stringable?

for example the user could have a clean strong typed enum and they would implement Display for debugging purposes.
If the user doesn't want to bother, they could use a Smallstr (a la monitoring UI?)

components/cu_transform/src/broadcaster.rs
```diff
@@ -0,0 +1,277 @@
...
+use cu29::prelude::*;
+use cu_spatial_payloads::Transform3D;
+use std::fmt::Debug;
+use std::marker::PhantomData;
+
+#[derive(Debug, Default, Clone)]
+pub struct TransformBroadcastPayload<
+    T: Copy + Debug + Default + bincode::enc::Encode + bincode::de::Decode<()> + 'static,
+> {
+    pub transforms: Vec<StampedTransform<T>>,
```
gbin, line 12: This needs to be constant size otherwise you will kill the copperlist linear access.

components/cu_transform/src/broadcaster.rs
```diff
@@ -0,0 +1,277 @@
...
+    where
+        Self: Sized,
+    {
+        Ok(Self {
+            _marker: PhantomData,
+        })
+    }
+
+    fn process(&mut self, clock: &RobotClock, new_msg: Self::Output) -> CuResult<()> {
+        new_msg.metadata.tov = Tov::Time(clock.now());
```
gbin, line 119: yeah, isn't it your time in your Stamped version?
If needed we might want to add a Set of TS in ToV:
  `Range(CuTimeRange)` <- I think this is the best we can do today but it could makes sense to add `Discreet(Array<CuTime>)` for that purpose

components/cu_transform/src/transform.rs
```diff
@@ -0,0 +1,371 @@
...
+/// Internal transform buffer that holds ordered transforms between two frames
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct TransformBufferInternal<T: Copy + Debug + 'static> {
+    transforms: VecDeque<StampedTransform<T>>,
+    max_capacity: usize,
+}
+
+/// Thread-safe wrapper around a transform buffer with concurrent access
+#[derive(Clone)]
+pub struct TransformBuffer<T: Copy + Debug + 'static> {
```
gbin, line 28: can we make that constant size too? because all those objects are really small, few bytes, we could have them in the stack, in the state of a task etc...

gbin:
Ok awesome first draft thanks a lot!

my main thing is that it is very ROS like and not Copper like :)

Everything is allocated, dynamic, stringly typed.

If you have buffers: introduce an N static size at compile time, make all the types in message *in* message with a Sized type, make all the task members Sized type *in* the task.

Let's make a rule: *any* allocation should be 1MB+ at the very minimum, native if as a task member, always a CuHandle if in messages.

Here everything is small a transform chain is like max like 10 element in a crazy snake like robot maybe? those are bytes, not megabytes -> everything should be Sized and copied.

components/cu_transform/src/velocity_cache.rs
```diff
@@ -0,0 +1,160 @@
...
+        &self,
+        from: &str,
+        to: &str,
+        time: CuTime,
+        path_hash: u64,
+    ) -> Option<VelocityTransform<T>> {
+        let key = (from.to_string(), to.to_string());
+
+        if let Some(mut entry) = self.entries.get_mut(&key) {
+            let now = Instant::now();
```
gbin, line 67: This cannot happen because of determinism. It has to be based on the robot clock otherwise the replay will be totally random. (imagine someone might go back in time, restore the state of the task and play again)

components/cu_transform/src/velocity.rs
```diff
@@ -0,0 +1,357 @@
...
+/// invalidated when new transforms are added or when entries exceed their age limit.
+#[derive(Debug, Clone, Encode, Decode, Serialize, Deserialize)]
+pub struct VelocityTransform<T: Copy + Debug + 'static> {
+    /// Linear velocity components [vx, vy, vz] in meters per second
+    pub linear: [T; 3],
+    /// Angular velocity components [wx, wy, wz] in radians per second
+    pub angular: [T; 3],
+}
+
+impl VelocityTransform<f64> {
```
gbin, line 48: maybe use https://docs.rs/num-traits/latest/num_traits and parametrize over all the compatible number formats?

components/cu_transform/src/tree.rs
```diff
@@ -0,0 +1,1572 @@
...
+
+/// The cache entry for a transform query
+#[derive(Clone)]
+struct TransformCacheEntry<T: Copy + Debug + 'static> {
+    /// The cached transform result
+    transform: Transform3D<T>,
+    /// The time for which this transform was calculated
+    time: CuTime,
+    /// When this cache entry was last accessed
+    last_access: Instant,
```
gbin, line 41: needs to be based on robotclock as we can restore tasks and go back in time and the determinism should be preserved

gbin:
moved the branch to this repo so we can collaborate easier on it.

PR #354, opened 2025-05-29
[POC] Tryout starting with cu-traits for no-std

gbin:
In order to measure the impact on the codebase, I started to experiment the port to no-std of in the order of `cargo workspaces list` (ie. the dependency tree order).

### DIFF
--- a/core/cu29_traits/src/lib.rs
+++ b/core/cu29_traits/src/lib.rs
@@ -1,20 +1,69 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+extern crate alloc;
+
+use alloc::string::ToString;
+use arrayvec::ArrayString;
 use bincode::{Decode as dDecode, Encode, Encode as dEncode};
+use core::error::Error;
+use core::fmt::{Debug, Display, Formatter, Result};
+use core::result;
 use serde::{Deserialize, Serialize};
-use std::error::Error;
-use std::fmt::{Debug, Display, Formatter};
+
+/// Most structures in Copper are fixed size. This is a wrapper for the strings.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CuString<const N: usize>(pub ArrayString<N>);
+
+impl<const N: usize> CuString<N> {
+    fn new() -> Self {
+        Self(ArrayString::new())
+    }
+}
+
+pub type CuStr16 = CuString<16>;
+pub type CuStr32 = CuString<32>;
+pub type CuStr64 = CuString<64>;
+pub type CuStr128 = CuString<128>;
+
+impl From<&str> for CuStr64 {
+    /// This will be a safe operation by truncating to the capacity
+    fn from(s: &str) -> Self {
+        let mut out = ArrayString::new();
+        out.push_str(&s[..s.len().min(64)]);
+        CuString::<64>(out.into())
+    }
+}
+
+#[cfg(feature = "std")]
+impl From<String> for CuStr64 {
+    fn from(value: String) -> Self {
+        value.as_str().into()
+    }
+}
+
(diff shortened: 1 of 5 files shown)


core/cu29_traits/src/lib.rs
```diff
@@ -1,20 +1,70 @@
...
+}
+
+pub type CuStr16 = CuString<16>;
+pub type CuStr32 = CuString<32>;
+pub type CuStr64 = CuString<64>;
+pub type CuStr128 = CuString<128>;
+
+impl From<&str> for CuStr64 {
+    /// This will be a safe operation by truncating to the capacity
+    fn from(s: &str) -> Self {
```
mik90, line 30: It'd be good to implement some `new_truncated` utility that tells you how many characters were truncated. The caller can definitely just check the length before making a fixed size string, but it can get kind of tedious

core/cu29_traits/src/lib.rs
```diff
@@ -1,20 +1,70 @@
...
+        Self(ArrayString::new())
+    }
+}
+
+pub type CuStr16 = CuString<16>;
+pub type CuStr32 = CuString<32>;
+pub type CuStr64 = CuString<64>;
+pub type CuStr128 = CuString<128>;
+
+impl From<&str> for CuStr64 {
```
mik90, line 28: Is it possible to impl a trait for the generic CuString type?

mik90: This compiles for me locally and passes a little unit test, although I have to delete the `From<&str> CuStr64` impl

```rust
impl<const N: usize> From<&str> for CuString<N> {
    /// This will be a safe operation by truncating to the capacity
    fn from(s: &str) -> Self {
        let mut out = ArrayString::new();
        out.push_str(&s[..s.len().min(N)]);
        CuString::<N>(out)
    }
}
```

Ditto for display

```rust
impl<const N: usize> Display for CuString<N> {
    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
        write!(f, "{}", self.0)
    }
}
```

mik90: Here's a branch https://github.com/copper-project/copper-rs/compare/gbin/cu29-traits/no-std...mkaliman/cu29-traits/no-std-suggestion

also added that truncation helper

mik90:
It's worth noting that clippy will warn on the return types here

```
warning: the `Err`-variant returned from this function is very large
   --> core/cu29_traits/src/lib.rs:136:35
    |
136 |     fn log(&mut self, obj: &E) -> CuResult<()>;
    |                                   ^^^^^^^^^^^^ the `Err`-variant is at least 140 bytes
    |
    = help: try reducing the size of `CuError`, for example by boxing large elements or replacing it with `Box<CuError>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
    = note: `#[warn(clippy::result_large_err)]` on by default
```

The types are kind of large, so i get it. But without an allocator and the ability to box things, what are we to do?

Another approach would be to do string interning like you did with `debug!` and allow that to handle the static components of errors. Still, there needs to be some level of dynamic strings to allow for things that we can't handle statically.

An enum-focused error API would avoid blowing up the stack at the cost of debuggability. Still, it would be possible to have different enums compose other base level enums so that you still get a tree of errors that can build up. And then `debug!` could be used for verbose error information that doesn't need to be handled at runtime. That does shift more of the debugging workflow to the unified logger, though

gbin:
Having slept on this, I think the other implementation is way more "Coppery", integrating the error messages to the interned string will benefit also the embedded version of Copper by reducing the binary size. 

Also this has a fatal flaw of increasing *a lot* all the return sizes on the stack.

Let's close this, thanks for the feedback.

PR #370, opened 2025-06-28
[Experimentation] Schema trait and implementation

gbin:
Context: when exporting data with cu29-export, we completely loose the structure of the output because internally we use bincode (which is schemaless)

This PR introduces a trait payloads can implement to allow this introspection of types.


gbin:
let's close that for now, maybe we could do that in a future version

PR #407, opened 2025-07-29
Adds a way to disable logging for some tasks.

gbin:
1. Added a logging { enabled = false } entry to the ron file
2. Generates the payload clears of the matching tasks on the runtime generation


gbin:
@matthewashton-k do you mind testing that on your project too?

PR #408, opened 2025-07-29
Matthewashton k/exclude null bytes from cucompactstring

matthewashton-k:


core/cu29_traits/src/lib.rs
```diff
@@ -140,8 +140,13 @@ pub struct CuCompactString(pub CompactString);
 impl Encode for CuCompactString {
     fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
         let CuCompactString(ref compact_string) = self;
-        let bytes = &compact_string.as_bytes();
-        bytes.encode(encoder)
+        let bytes = compact_string.as_bytes();
+        let filtered_len = bytes.iter().filter(|&&b| b != 0).count();
+        filtered_len.encode(encoder)?;
+        for &byte in bytes.iter().filter(|&&b| b != 0) {
+            byte.encode(encoder)?;
```
matthewashton-k, line 147: There's probably a cleaner way to do this but this is what I came up with that doesn't need to allocate a new Vec

gbin:
fixed at the root of the problem with #409 thanks!

PR #416, opened 2025-08-01
Fix threadpool use after move.

matthewashton-k:


examples/cu_background_task/copperconfig.ron
```diff
@@ -6,16 +6,22 @@
         ),
         (
             id: "task1",
-            type: "tasks::ExampleTask",
+            type: "tasks::ExampleTaskSlower",
             background: true,
         ),
         (
             id: "task2",
+            type: "tasks::ExampleTaskSlow",
```
gbin: Maybe reuse the same type but add a wait parameter?

gbin:
Tip: when opening a PR in the title explain the *why* and the what: ie. We can see you added a clone in the code but the collaborators might not know why.
Thanks!

examples/cu_background_task/copperconfig.ron
matthewashton-k: Sure thing!

PR #422, opened 2025-08-14
Adding an example that breaks on multi inputs

gbin:


matthewashton-k:
Seems to be that the problem is that CuSimSinkTask only supports one input:
```rust
impl<T: CuMsgPayload> CuSinkTask for CuSimSinkTask<T> {
    type Input<'m> = input_msg!(T);

    fn new(_config: Option<&ComponentConfig>) -> CuResult<Self>
    where
        Self: Sized,
    {
        Ok(Self { boo: PhantomData })
    }

    fn process(&mut self, _clock: &RobotClock, _input: &Self::Input<'_>) -> CuResult<()> {
        unimplemented!("A placeholder for sim was called for a sink, you need answer SimOverride to ExecutedBySim for the Process step.")
    }
}
```

gbin:
@matthewashton-k sorry I took so long to come back to this ... Hopefully this fix works, it is forwarding the input tuple for the placeholder sim sink through templating.

PR #439, opened 2025-09-25
Balancebot example integration test and miscellaneous changes

henrygerardmoore:
I've been starting to learn copper recently and I really like the balancebot example! In going through it I found a few changes that I think improve the example. Broadly, here are the changes I made:

- Added an integration test that just creates the app and updates it once, running headlessly
- Miscellaneous comment changes and additions
- Consolidated the logic for dragging the cart and the rod into one place, separated into one observer for the drag and one for the drag ending. My changes seemed to make the dragging a little smoother since it lets avian handle the position updates instead of directly modifying the transform.
- A few camera control changes, making the camera use real time instead of not specifying (I think `Time` defaults to `Time<Virtual>` where it was being used, meaning e.g. zooming the camera didn't work when time was paused. Also make the camera always face the origin after a move, whereas before it would jump back to facing the origin when the middle mouse was pressed
- Added an optional gizmo to display the external forces acting on the system. I just think it's cool to see the cart and be able to see the relative magnitude of your dragging force

Feel free to ask me to remove any or all of this! I mostly made it in hopes it would be helpful and to actively learn via the example.

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -30,6 +30,8 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustc,cargo,rust-std,rustfmt,clippy
       - name: Setup rust-cache
         uses: Swatinem/rust-cache@v2
         with:
--- a/examples/cu_rp_balancebot/src/sim.rs
+++ b/examples/cu_rp_balancebot/src/sim.rs
@@ -209,6 +209,7 @@ pub fn make_world(headless: bool) -> App {
     if headless {
         // these are not added in the minimal plugins but are needed for the integration test to run
         world.insert_resource(Assets::<Mesh>::default());
+        world.insert_resource(Assets::<Font>::default());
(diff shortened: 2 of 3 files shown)


henrygerardmoore:
Ah looks like from CI that macOS will need another asset added in headless. I'll add that. Unfortunately I don’t have a mac so I'll rely on CI to test it. I haven’t yet tested in release mode locally but I'll do that, too, and fix both those problems.

Then I can split out the PR. I think I'll split it into two PRs, one for the integration test with no functional changes and then one with features that points to that branch. Does that sound good?

henrygerardmoore:
@makeecat I can drag when built in release (on ubuntu 22.04, haven't yet tried on 24.04 but I will later today), could it be that the simulation is paused? I did change it so you cannot drag while simulation is paused in this PR. If that's the case, perhaps it should be more obvious when the simulation is paused? Or I can make it so that you can drag while the simulation is paused, though it won't really do anything but I guess display a force with the gizmo.

henrygerardmoore:
Closing to split into multiple PRs, first of which is #440

PR #441, opened 2025-09-26
Balancebot stability improvement and feature addition

henrygerardmoore:
These are the changes from #439 other than the integration test and comment changes, namely:

- Consolidated the logic for dragging the cart and the rod into one place, separated into one observer for the drag and one for the drag ending. My changes seemed to make the dragging a little smoother since it lets avian handle the position updates instead of directly modifying the transform. It also lets you 'hold' a drag since the previous logic only applied the transform update when the mouse was moved while dragging.
- A few camera control changes, making the camera use real time instead of not specifying (I think `Time` defaults to `Time<Virtual>` where it was being used, meaning e.g. zooming the camera didn't work when time was paused). Also make the camera always face the origin after a move, whereas before it would jump back to facing the origin when the middle mouse was pressed
- Added an optional gizmo (toggled with the F key) to display the external forces acting on the system. I just think it's cool to see the cart compensating and be able to see the relative magnitude of your dragging force

image
Here's a picture of the force gizmo in action. I'm dragging the rod up and left with my mouse, and the cart was moving right when I captured the screenshot

I tested on release and debug on Ubuntu 22.04 and 24.04 and everything seems to work

### DIFF
--- a/examples/cu_rp_balancebot/src/world/mod.rs
+++ b/examples/cu_rp_balancebot/src/world/mod.rs
@@ -483,23 +483,66 @@ fn setup_entities(
     setup_completed.0 = true; // Mark as completed
 }
 
+fn get_rigid_body_entity(
+    mut drag_target: Entity,
+    parents: &Query<(&ChildOf, Option<&RigidBody>)>,
+) -> Entity {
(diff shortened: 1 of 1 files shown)


gbin (approved):
Nice thanks

PR #488, opened 2025-11-07
Fix README links to point to GitHub wiki pages instead of doc

ahmedkhalaf:
Links pointing to doc/*.md are broken, replaced with equivalent pages under https://github.com/copper-project/copper-rs/wiki/


gbin:
thanks

PR #491, opened 2025-11-07
Pass on an API for the revised CuBridge design

gbin:
a little wonky because trait specializations are still experimental in Rust.

We might want to change it to send<Id, MsgType> once it is implemented so users would just need to add an implementation of each of the valid pairs with 0 dispatching.


kamibo (approved):
LGTM

PR #497, opened 2025-11-11
Bidirectional Dshot implemented as a Cubridge + demo

gbin:
image

End to end with logging, etc...

### DIFF
--- a/components/bridges/cu_bdshot/src/board.rs
+++ b/components/bridges/cu_bdshot/src/board.rs
@@ -125,6 +125,7 @@ impl BdshotBoard for Rp2350Board {
 }
 
 impl Rp2350Board {
+    /// Basic DSHOT transaction, sending a frame and attempting to read a telemetry response.
     fn exchange_impl<SM>(
         cycles_per_micro: u32,
         ch: &mut EscChannel<pac::PIO0, SM, Word, Word>,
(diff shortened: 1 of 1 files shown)


.github/workflows/general.yml
makeecat, line 1: Can you update them in the weekly.yml as well?

Cargo.toml
```diff
@@ -167,3 +170,9 @@ panic = "abort"
 strip = "symbols"  # or "debuginfo" or true
 debug = false
 incremental = false
+
+[profile.dev]
+opt-level = 0
+debug = 2
+debug-assertions = true
+overflow-checks = true
```
makeecat: should we have a separate PR for cargo profile.dev?

components/bridges/cu_bdshot/src/board.rs
```diff
@@ -0,0 +1,205 @@
...
+    !((d << 4) | crc) as u32
+}
+
+pub fn encode_frame(command: EscCommand) -> u32 {
+    let mut bits = (command.throttle & 0x07FF) << 1;
+    if command.request_telemetry {
+        bits |= 0x01;
+    }
+    bdshot_frame(bits)
+}
```
makeecat, line 189: need comments

makeecat (changes requested):
looks good to me.
we may need to come back to do plumbing for CI later. let's merge this first.

.github/workflows/general.yml
gbin: as per comment let's not bring too much into this already totally bloated PR. Thanks

PR #499, opened 2025-11-13
Moved the route parameter on CuBridge to optional

gbin:
lots of bridges do have a natural ID -> Route map and suddenly this mandatory route parameted make them stutturing

for example an AIO ESC have 4 motors, you cannot route anything but the API was forcing you to add a random string to it.

This will be followed by some helper to generate more complex bridges like middleware ones where the user needs to generate it at the last minute with all the types they want to expose to each ID.

### DIFF
--- a/core/cu29_runtime/src/cubridge.rs
+++ b/core/cu29_runtime/src/cubridge.rs
@@ -17,6 +17,7 @@ use std::borrow::Cow;
 #[cfg(not(feature = "std"))]
 mod imp {
     pub use alloc::fmt::{Debug, Formatter};
+    pub use alloc::string::String;
 }
 
 #[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


core/cu29_runtime/src/cubridge.rs
```diff
@@ -22,33 +28,49 @@ use imp::*;
...
+    pub const fn with_channel(id: Id, route: &'static str) -> Self {
+        Self {
+            id,
+            default_route: Some(route),
             _payload: PhantomData,
         }
     }
+
+    /// Backward-compatibility helper for older code; prefer [`BridgeChannel::with_channel`].
+    #[deprecated(since = "0.11.0", note = "Use BridgeChannel::with_channel instead")]
```
gbin: yolo ai coded, this needs to go

PR #511, opened 2025-11-19
Vendored tui-nodes and bent it to our will for bridges.

gbin:
Some yolo dev to render correctly the bridges in the TUI DAG

image

image


gbin:
(the screenshot is already stale, it looks even better)
image

PR #535, opened 2025-11-27
Temporary EOF Log Marker

ishanchadha01:
- includes UnifiedLogger lib definition
- implementation of temporary EOF marker that's overwritten when new log entry is written
- tests for the temporary marker and permanent marker
- mirrored functionality in SDLogger

### DIFF
--- a/examples/cu_elrs_bdshot_demo/src/main.rs
+++ b/examples/cu_elrs_bdshot_demo/src/main.rs
@@ -18,9 +18,9 @@ use embedded_sdmmc::SdCard;
 use rp235x_hal as hal;
 use rp235x_hal::clocks::Clock;
 use rp235x_hal::fugit::RateExtU32;
-use rp235x_hal::gpio::bank0::{Gpio12, Gpio13, Gpio15, Gpio16, Gpio17, Gpio18, Gpio19};
+use rp235x_hal::gpio::bank0::{Gpio15, Gpio16, Gpio17, Gpio18, Gpio19, Gpio2, Gpio3};
 use rp235x_hal::gpio::{
-    Function, FunctionPio0, FunctionSio, FunctionSpi, FunctionUart, FunctionXipCs1, Pin, PinId,
+    Function, FunctionPio0, FunctionSio, FunctionSpi, FunctionUartAux, FunctionXipCs1, Pin, PinId,
     PullDown, PullNone, PullType, PullUp, SioInput, SioOutput, ValidFunction,
 };
 use rp235x_hal::pac::{SPI0, UART0};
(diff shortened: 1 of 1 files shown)


gbin:
`that's overwritten when new log entry is written` <-  more like new section is written?

components/libs/cu_sdlogger/src/logger.rs
```diff
@@ -226,6 +228,63 @@ impl<BD: BlockDevice> EMMCLogger<BD> {
...
+        if end_block > self.last_block {
+            return Err(CuError::from("out of space"));
+        }
+
+        let header = SectionHeader {
+            magic: SECTION_MAGIC,
+            block_size: SECTION_HEADER_COMPACT_SIZE,
+            entry_type: UnifiedLogType::LastEntry,
+            offset_to_next_section: (blocks_needed * block_size) as u32,
+            used: payload_len as u32,
```
gbin: there is no payload on a LastEntry, should be always 0.

core/cu29_unifiedlog/src/memmap.rs
```diff
@@ -192,6 +193,65 @@ impl SlabEntry {
...
+        let marker_end = marker_start + total_marker_size;
+        if marker_end > self.mmap_buffer.len() {
+            return Err("Not enough space to write end-of-log marker".into());
+        }
+
+        let header = SectionHeader {
+            magic: SECTION_MAGIC,
+            block_size: SECTION_HEADER_COMPACT_SIZE,
+            entry_type: UnifiedLogType::LastEntry,
+            offset_to_next_section: total_marker_size as u32,
```
gbin: same thing here there is no need to compute that

core/cu29_unifiedlog/src/memmap.rs
```diff
@@ -192,6 +193,65 @@ impl SlabEntry {
...
+        let total_marker_size = block_size + marker_size;
+        let marker_end = marker_start + total_marker_size;
+        if marker_end > self.mmap_buffer.len() {
+            return Err("Not enough space to write end-of-log marker".into());
+        }
+
+        let header = SectionHeader {
+            magic: SECTION_MAGIC,
+            block_size: SECTION_HEADER_COMPACT_SIZE,
+            entry_type: UnifiedLogType::LastEntry,
```
gbin, line 217: Writing a new payload for basically a boolean is overkill.


I'd like to propose an alternative that might be better: as we write all the section header when we close them, we could add a "still open" boolean in all the section headers. The first time we create them we leave the boolean at open, then when we close them we flip the boolean (we write that sector anyway so it is almost free).

We use this flag on the temporary end marker to signal that as a while the entire log was not closed properly.

Like that the log reader can detect the still open / dirty sections when the log has not been closed properly. 

Then we need to adapt the log reader and fsck to emit a warning that they stopped reading the log on a not cleanly closed log and list the sections that were still open (hinting to the user that they miss some data in there).

components/libs/cu_sdlogger/src/logger.rs
```diff
@@ -226,6 +228,63 @@ impl<BD: BlockDevice> EMMCLogger<BD> {
...
+    fn clear_temporary_end_marker(&mut self) {
+        if let Some(marker) = self.temporary_end_marker.take() {
+            self.next_block = marker;
+        }
+    }
+
+    fn write_end_marker(&mut self, temporary: bool) -> CuResult<()> {
+        let marker = EndOfLogMarker { temporary };
+
+        let mut payload = [0u8; 32];
```
gbin: I think this is overkill to create a payload for this. See below

gbin:
thanks for the PR, see inside but I think we would be better off if we add this boolean on every section header see the comments in the review.

PR #584, opened 2025-12-28
Cleanup the main justfile now:

gbin:
just (default/no parameter)-> runs std-ci
std-ci -> runs local something close to CI runs for std nostd-ci -> runs local something close to CI runs for nostd ci -> runs the FULL CI with act and docker.


gbin:
needs #582

PR #610, opened 2026-01-06
workspace template

makeecat:
address issue: https://github.com/copper-project/copper-rs/issues/142


templates/README.md
```diff
@@ -62,15 +62,22 @@ You can generate a new Copper project using the `cargo generate` command along w
...
 
-The generated project includes helpful utility scripts in its `bin/` directory:
+The generated project includes helpful utility scripts in its `scripts/` directory:
 
-* `bin/log`: Use this script to extract and view structured logs generated by your Copper application.
-* `bin/cl`: Use this script to extract CopperLists (sequences of operations) from the application's log output.
-* `bin/rcfg`: Use this script to generate a `.dot` file representing the application's execution Directed Acyclic Graph (DAG). You can use tools like Graphviz to visualize this file (e.g., `dot -Tpng graph.dot -o graph.png`).
+* `scripts/log`: Use this script to extract and view structured logs generated by your Copper application.
+* `scripts/cl`: Use this script to extract CopperLists (sequences of operations) from the application's log output.
+* `scripts/rcfg`: Use this script to generate a `.dot` file representing the application's execution Directed Acyclic Graph (DAG). You can use tools like Graphviz to visualize this file (e.g., `dot -Tpng graph.dot -o graph.png`).
```
gbin: Should we move that to justfiles like we did for us?

templates/cu_full/cargo-generate.toml
```diff
@@ -2,9 +2,9 @@
 pre = ["init.rhai"]
 
 [template]
-description = "A template for a Copper-based Rust project"
+description = "A template for a Copper-based Rust workspace"
```
gbin: I think we should have both ie. project and workspace.

templates/cu_full/README.md
```diff
@@ -0,0 +1,19 @@
+# {{project-name}} Workspace
+
+This template bootstraps a multi-crate Copper workspace intended to scale beyond a single binary.
+
+## Layout
+
+- `apps/{{project-name}}/`: runtime binary, log reader, and config.
+- `tasks/{{project-name}}-tasks/`: Copper task implementations.
```
gbin: I think we can improve this structure:
for cu_full the goal is really to have a workspace a little bit like the copper-rs repo where the users will have several apps (instead of examples) with a set of tasks & messages that are shared with across those apps.
So tasks/{{project-name}}-tasks/ doesn't make much sense: if those are tasks for only *that* apps/{{project-name}}/ then it will be in apps/{{project-name}}/, it would be better to provide a workspace with let's say 2 dummy apps, 3 dummy tasks and a custom message implemented. That would help people getting started on the right foot.

templates/cu_full/README.md
```diff
@@ -0,0 +1,19 @@
...
+- `libs/{{project-name}}-types/`: shared message types.
```
gbin: why not mimicking the copper-rs hierarchy so people have less cognitive overload to learn from 2 different structures? https://github.com/copper-project/copper-rs/tree/master/components

Also it will help when people will start to share their own projects.

templates/cu_full/apps/{{project-name}}-demo/config/copperconfig.ron
```diff
@@ -0,0 +1,27 @@
+// A sample configuration to get you started.
+(
+    tasks: [
+        (
+            id: "src",
+            type: "{{project-name|snake_case}}_source::MySource",
```
gbin: This is still strange to me here: why make a sub module with project-name? If you have tasks *specific* to one project... They should be in the app not shared in components no? If they are shared then it is "MySource" with its crate my-source (or something like that) which is in the components/sources

gbin (changes requested):

gbin:
I would seed the workspace template generating something like that, with the simpler application template just generating the one with *.

The goal is that it is runnable immediately after generation of course, logging at the right place etc..

```
├── Cargo.toml
├── justfile                
├── .gitignore               // for target etc...
├── README.md
├── apps
│   ├── README.md
│   └── cu_example_app                     *
│       ├── Cargo.toml                     *
│       ├── README.md                      *
│       ├── copperconfig.ron               *
│       ├── src                            *
│       │   ├── main.rs                    *
│       │   ├── messages.rs                *
│       │   └── tasks                      *
│       │       ├── mod.rs                 *
│       │       ├── local_example_src.rs   *
│       │       ├── local_example_task.rs  *
│       │       └── local_example_sink.rs  *
│       └── logs
│           └── .keep                // make it log there by default.
├── components
│   ├── README.md                    // state this is for shared components. 
│   ├── bridges
│   │   ├── .keep
│   │   └── cu_example_shared_bridge
│   │       ├── Cargo.toml
│   │       ├── README.md
│   │       └── src
│   │           ├── lib.rs
│   │           └── messages.rs
│   ├── monitors    
│   │   └── .keep
│   ├── payloads
│   │   └── .keep
│   ├── sinks
│   │   └── .keep
│   ├── sources
│   │   └── .keep
│   └── tasks
│       └── .keep
└── doc
    └── .keep
```

gbin (approved):

Issue #231, opened 2025-01-22
Test Copper with RTSan


gbin:
The LLVM RTSan is now available on experimental Rust:
https://steck.tech/posts/rtsan-in-rust/

Validating real-timeness for at least an execution through a CopperList might be interesting

makeecat:
addressed in #612.
TODOs: add documentation in the Wikis

PR #612, opened 2026-01-06
rtsan example for caterpillar

makeecat:
## Summary
Added a rtsan-smoke just target and exposed an rtsan feature in the cu-caterpillar
example to pass through cu29/rtsan, so the target can actually enable RTSan on a runnable
app.
Related Issue: https://github.com/copper-project/copper-rs/issues/231
## Details
- justfile: new rtsan-smoke recipe with params (pkg, bin, args, options) that runs cargo
  run with RTSAN_ENABLE=1 and RTSAN_OPTIONS.
- examples/cu_caterpillar/Cargo.toml: added rtsan = ["cu29/rtsan"] so --features rtsan
  works on the example.
- run `just rtsan-smoke` for the example. (It should fail)


core/cu29_derive/src/lib.rs
```diff
@@ -1106,7 +1110,11 @@ pub fn copper_runtime(args: TokenStream, input: TokenStream) -> TokenStream {
                         quote! {
                             #call_sim_callback
                             if doit {
-                                if let Err(error) = tasks.#task_index.postprocess(clock) {
+                                let maybe_error = {
+                                    let _rt_guard = ::cu29::rtsan::ScopedSanitizeRealtime::default();
```
gbin: maybe we can define an rt_guard expression if the rtsan feature flag is on and inject it with #rt_guard. It would avoid the dummy implementation at the beginning

gbin: how is it reporting the failure btw? I don't see any

makeecat: If you run just rtsan-smoke, you should see the error message

Issue #465, opened 2025-10-06
Investigate openZL for copper logs


gbin:
https://engineering.fb.com/2025/10/06/developer-tools/openzl-open-source-format-aware-compression-framework/

It looks like a good optimization for a compressed version of Copper logs.

PR #614, opened 2026-01-06
test openzl for copper log compression

makeecat:
## Summary
Offline benchmark tool for Copper log archival evaluation. This initial version targets the
balancebot schema defined in `examples/cu_rp_balancebot/copperconfig.ron` and emits columnar
streams plus a `manifest.json` with zstd baselines.
Related Issue: https://github.com/copper-project/copper-rs/issues/465

### DIFF
--- a/support/cu29_archive_bench/scripts/openzl_eval.py
+++ b/support/cu29_archive_bench/scripts/openzl_eval.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+"""Run OpenZL compression per columnar stream using profile mapping."""
+
+import argparse
+import json
+import subprocess
+from pathlib import Path
+
+PROFILE_MAP = {
+    ".u16": "le-u16",
+    ".u32": "le-u32",
+    ".u64": "le-u64",
+    ".i16": "le-i16",
+    ".i32": "le-i32",
+    ".i64": "le-i64",
+}
+
+SERIAL_SUFFIXES = [
+    ".u8",
+    ".i8",
+    ".f32",
+    ".f64",
+    ".bytes",
+]
+
+
+def choose_profile(stream_name: str) -> str:
+    for suffix, profile in PROFILE_MAP.items():
+        if stream_name.endswith(suffix):
+            return profile
+    for suffix in SERIAL_SUFFIXES:
+        if stream_name.endswith(suffix):
+            return "serial"
+    return "serial"
+
+
+def run(cmd):
+    res = subprocess.run(cmd, capture_output=True, text=True)
+    return res.returncode, res.stdout.strip(), res.stderr.strip()
+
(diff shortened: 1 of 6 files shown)


support/cu29_archive_bench/scripts/openzl_eval.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
```
gbin, line 1: Usually for one shot stuff/test things I try to put them in a separate repo to minimize the amount of code to actually maintain: see https://github.com/copper-project/stm32_bmi088 where I tested first the bmi088 before integrating it to the main repo. I can create one for you if you need.

support/cu29_archive_bench/scripts/openzl_train_eval.py
```diff
@@ -0,0 +1,251 @@
...
+    compress_dir = Path(args.compress_dir) if args.compress_dir else base_dir / "openzl_trained_compressed"
+    result_json = Path(args.result_json) if args.result_json else base_dir / "openzl_trained_results.json"
+
+    train_dir.mkdir(parents=True, exist_ok=True)
+    compress_dir.mkdir(parents=True, exist_ok=True)
+
+    results = []
+    errors = []
+
+    for file_path, rel, stream_name in iter_streams(base_dir, structured_dir, copper_dir):
```
gbin, line 95: We need to keep in mind the actual goal here is not compression only.

Compressed data means less precious bandwidth towards the storage BUT needs more CPU before it can send it to disk.

So what we want to measure is basically this tradeoff:

1. for a given CPU speed
2. with the data already hot in L2/L3 cache (imagine the copperlist just executed so it is sitting there)
3. and a given storage speed 

We need to know:
1. can we recoup the CPU cost with the storage gain?
2. what would then be the storage space gain

Ideally for 2 profiles of machines we target: one SBC with an nvme drive and one MPU with an SDCard. No need to really test yet on real HW but we can make educated guesses from the specs with a known baseline if we know the number of retired instruction per pass + nb of MB to write, you can know that with linux perf for example.

support/cu29_archive_bench/scripts/openzl_eval.py
makeecat: Do you want to remove the entire crates outside of the mono-repo? If yes, feel free to do so. I created this PR for investigating OpenZL instead of adding a feature. We should use a separate repo for this purpose.

support/cu29_archive_bench/scripts/openzl_train_eval.py
makeecat: That makes sense, once you create the repo for investigation of OpenZL, we can conduct these profiles that matter.

Issue #191, opened 2024-12-31
Confirm cu-aligners work with memory buffers


gbin:
https://github.com/copper-project/copper-rs/tree/master/components/tasks/cu_aligner

This task should benefit a lot from the new memory buffers defined in core/pool.rs.

Simply make an example and confirms it allows aligning images in time for example

PR #615, opened 2026-01-06
propose cu-image-aligner example

makeecat:
## Summary

Adds a cu-aligner image alignment example using pooled image buffers, with minor safety and payload handling improvements in cu-aligner and workspace dep cleanup.

## Changes

- Add new cu_image_aligner example app, config, tasks, and README under `examples/cu_image_aligner/`
- Guard against time underflow in cu-aligner buffer horizon/window calculations in `components/tasks/cu_aligner/src/buffers.rs`
- Initialize output payload once and clone payloads to support non-Copy types, plus add string payload test in `components/tasks/cu_aligner/src/lib.rs`
- Move paste to workspace dependencies and align usage in `Cargo.toml`, `components/tasks/cu_aligner/Cargo.toml`, and `core/cu29_runtime/Cargo.toml`

Related Issue: https://github.com/copper-project/copper-rs/issues/191

### DIFF
--- a/components/tasks/cu_aligner/src/buffers.rs
+++ b/components/tasks/cu_aligner/src/buffers.rs
@@ -120,11 +120,8 @@ macro_rules! alignment_buffers {
             /// Call this to be sure we discard the old/ non relevant data
             #[allow(dead_code)]
             pub fn purge(&mut self, now: cu29::clock::CuTime) {
-                let horizon_time = if now >= self.stale_data_horizon {
-                    now - self.stale_data_horizon
-                } else {
-                    cu29::clock::CuDuration::MIN
-                };
+                use cu29::prelude::SaturatingSub;
+                let horizon_time = now.saturating_sub(self.stale_data_horizon);
                 // purge all the stale data from the TimeboundCircularBuffers first
                 $(self.$name.purge(horizon_time);)*
             }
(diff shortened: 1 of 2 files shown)


components/tasks/cu_aligner/src/buffers.rs
```diff
@@ -120,7 +120,11 @@ macro_rules! alignment_buffers {
             /// Call this to be sure we discard the old/ non relevant data
             #[allow(dead_code)]
             pub fn purge(&mut self, now: cu29::clock::CuTime) {
-                let horizon_time = now - self.stale_data_horizon;
+                let horizon_time = if now >= self.stale_data_horizon {
+                    now - self.stale_data_horizon
+                } else {
```
gbin: is this simply a saturation_sub? 
```rust
let horizon_time = now.saturating_sub(self.stale_data_horizon);
```

we could add the trait to CuTime if it does not exist.

components/tasks/cu_aligner/src/buffers.rs
```diff
@@ -144,7 +148,11 @@ macro_rules! alignment_buffers {
 
                 let most_recent_time = most_recent_time.unwrap();
 
-                let time_to_get_complete_window = most_recent_time - self.target_alignment_window;
+                let time_to_get_complete_window = if most_recent_time >= self.target_alignment_window {
```
gbin: same

components/tasks/cu_aligner/src/buffers.rs
makeecat: Yes, would it be more straightforward to add a helper function in `core/cu29_clock/src/lib.rs` instead of a trait?

gbin: I think this is super standard as a trait in Rust: https://docs.rs/num/latest/num/traits/trait.SaturatingSub.html

so it might help when we want to give times to other APIs that would need it.

gbin: add the trait is more *implement* the trait sorry

Issue #383, opened 2025-07-06
Implement multiple outputs per task after all


gbin:
This would be handled the same way as the multiple inputs through CuMsgPacks.

Ie. At each cycle we would consider a tuple of outputs from the tasks.

PR #635, opened 2026-01-12
feat: multi-output task support

makeecat:
## Summary

Implemented multi-output task support end-to-end (runtime plan, codegen, monitoring topology, sim support) with deterministic port ordering and backward compatibility

Related Issue: https://github.com/copper-project/copper-rs/issues/383

## Detail Changes

- Extended output_msg! to return tuple packs and updated runtime plan/output-port mapping logic for multi-output routing (`core/cu29_runtime/src/cutask.rs`, `core/cu29_runtime/src/curuntime.rs`).
- Updated monitoring topology and render topology to expose ordered output ports and keep metadata stable (`core/cu29_runtime/src/monitoring.rs`, `core/cu29_runtime/src/config.rs`).
- Reworked derive codegen for output packs, getters, sim support, stamping/clearing, and bridge routing (`core/cu29_derive/src/lib.rs`).
- Added multi-output example app and README documentation, plus workspace membership updates (`examples/cu_multi_output`, `Cargo.toml`, `README.md`).
- Builds inputs from incoming edge IDs and sorts by edge id so a task can receive multiple outputs from the same source without type-order mismatches
- Adds a TOV on the first output, includes the extra sinks/relay task, and updates LoopbackTask to a single-input pass-through.

### DIFF
--- a/core/cu29_runtime/src/curuntime.rs
+++ b/core/cu29_runtime/src/curuntime.rs
@@ -30,7 +30,7 @@ use cu29_log_runtime::log_debug_mode;
 use cu29_value::to_value;
 
 use alloc::boxed::Box;
-use alloc::collections::VecDeque;
+use alloc::collections::{BTreeSet, VecDeque};
 use alloc::format;
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
@@ -845,6 +845,32 @@ pub fn compute_runtime_plan(graph: &CuGraph) -> CuResult<CuExecutionLoop> {
                 queue.push_back(neighbor);
             }
         }
+    }
+
+    let mut planned_nodes = BTreeSet::new();
+    for unit in &plan {
+        if let CuExecutionUnit::Step(step) = unit {
+            planned_nodes.insert(step.node_id);
+        }
+    }
+
+    let mut missing = Vec::new();
+    for node_id in graph.node_ids() {
+        if !planned_nodes.contains(&node_id) {
+            if let Some(node) = graph.get_node(node_id) {
+                missing.push(node.get_id().to_string());
+            } else {
+                missing.push(format!("node_id_{node_id}"));
+            }
+        }
+    }
+
+    if !missing.is_empty() {
+        missing.sort();
+        return Err(CuError::from(format!(
+            "Execution plan could not include all nodes. Missing: {}. Check for loopback or missing source connections.",
+            missing.join(", ")
+        )));
     }
 
     Ok(CuExecutionLoop {
--- a/core/cu29_runtime/tests/loopback.rs
+++ b/core/cu29_runtime/tests/loopback.rs
@@ -1,0 +1,20 @@
+#[cfg(all(test, feature = "std"))]
+mod tests {
+    use cu29_runtime::config::read_configuration;
+    use cu29_runtime::curuntime::compute_runtime_plan;
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_loopback_config_rejected() {
+        let config_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("tests")
+            .join("loopback_config.ron");
+        let config =
+            read_configuration(config_path.to_str().unwrap()).expect("config should parse");
+        let graph = config.get_graph(None).expect("graph should load");
+        let err = compute_runtime_plan(graph).expect_err("loopback should fail");
+        let msg = err.to_string();
+        assert!(msg.contains("loopback"), "unexpected error: {msg}");
+        assert!(msg.contains("Missing"), "unexpected error: {msg}");
+    }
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -237,6 +236,27 @@ fn main() {
...
+    Ok(())
+}
+```
+
+Connections still use `(src, dst, msg)` only. Output ports are selected by `msg` type:
+- Each unique `msg` type in a task's outgoing `cnx` entries becomes one output port.
+- Port ordering follows the first appearance order in `cnx` for that source task.
+- Fan-out with the same `msg` type shares the same output port.
+
+Limitation: A task cannot expose two distinct output ports with the same payload type.
```
gbin: Giving it some thoughts but I think it might be reasonable as a quirk. The alternative would be named outputs like the bridges but this makes this super complicated and wonky on the backward compat.

README.md
```diff
@@ -237,6 +236,27 @@ fn main() {
 But this is a very minimal example for a task; please see [lifecycle](https://github.com/copper-project/copper-rs/wiki/Task-Lifecycle) for a more complete explanation
 of a task lifecycle.
 
+#### Multiple Outputs Per Task
```
gbin: This should be in the wiki, not the README

examples/cu_multi_output/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+            _resources: Self::Resources<'_>,
+        ) -> CuResult<Self>
+        where
+            Self: Sized,
+        {
+            Ok(Self {})
+        }
+
+        fn process(&mut self, _clock: &RobotClock, output: &mut Self::Output<'_>) -> CuResult<()> {
+            output.0.set_payload(42);
```
gbin, line 33: maybe add a TOV on one of the output as they are independent

examples/cu_multi_output/copperconfig.ron
```diff
@@ -0,0 +1,11 @@
+(
```
gbin, line 1: Can we do like the bridge testing here and add a bunch of missions with some variation in the routing like here https://github.com/copper-project/copper-rs/blob/master/examples/cu_bridge_test/copperconfig.ron because there are a lot of corner cases:

what if one task gets both inputs? if an output is left open? Is a loopback error is correctly spotted?

core/cu29_runtime/src/config.rs
```diff
@@ -1777,6 +1777,24 @@ pub(crate) fn build_render_topology(graph: &CuGraph, bridges: &[BridgeConfig]) -
...
+    output_edges.sort_by_key(|edge| edge.id().index());
+    for edge in output_edges {
+        let cnx = edge.weight();
+        if let Some(&idx) = node_lookup.get(&cnx.src)
+            && nodes[idx].flavor == Flavor::Task
+            && cnx.src_channel.is_none()
+        {
+            let port_map = &mut output_port_lookup[idx];
+            if !port_map.contains_key(&cnx.msg) {
+                let label = format!("out{}:{}", port_map.len(), cnx.msg);
```
gbin: add a space: `"out{}: {}"`

core/cu29_runtime/src/monitoring.rs
```diff
@@ -97,6 +97,31 @@ struct NodeIoUsage {
...
+        let Some(edge) = graph.edge(edge_id) else {
+            continue;
+        };
+        if seen.iter().any(|msg| msg == &edge.msg) {
+            continue;
+        }
+        seen.push(edge.msg.clone());
+        let mut port_label = String::from("out");
+        port_label.push_str(&port_idx.to_string());
+        port_label.push(':');
```
gbin: here too I guess for the space

examples/cu_multi_output/src/main.rs
makeecat: done in https://github.com/copper-project/copper-rs/pull/635/changes/1265de498c0b182862824fb862cb6f7748326128

examples/cu_multi_output/copperconfig.ron
makeecat: - Both inputs: covered by the both_inputs mission in `examples/cu_multi_output/copperconfig.ron` and
  `tasks::BothSink` in `examples/cu_multi_output/src/main.rs`.
- Output left open: covered by the open_output mission in `examples/cu_multi_output/copperconfig.ron` using
  `tasks::DropBoolSink` in `examples/cu_multi_output/src/main.rs` to intentionally drop one output.
- Loopback error: validated via a dedicated config+test (`core/cu29_runtime/tests/loopback_config.ron`, `core/
  cu29_runtime/tests/loopback.rs`) and explicit detection in `core/cu29_runtime/src/curuntime.rs`. The example
  config no longer includes a self-loop because it breaks compile-time planning.

Issue #120, opened 2024-12-02
Feature Request: log in debug! the copper list number 


gbin:
Today we have a structured log to add arbitrary debug information in the log. It would be nice if debug! can capture the context of Copper deterministic replay (for example the copper list number) and attach it to the logged info.

Like this the debug info can be reattached to the overall execution at the message level instead of just timestamp / line-number.

makeecat:
If Copper tasks can run in parallel, how to implement it? I can only think about a thread-local approach which assumes single-threaded execution of copper lists.

gbin:
They will need to be staggered with a unique CL number anyway.

gbin:
It is hard to do without some locking or sync, and this info is in every single logged copperlist so let's skip that, too much runtime constraints for little usability gain

PR #645, opened 2026-01-12
[FEAT] log in debug! the copper list number

makeecat:
## Summary

Copper structured logs now capture the current CopperList id automatically, with per-thread context tracking, backward-compatible decoding for older logs, and culistid surfaced in Rust/Python readers plus tests and docs updates.

Related Issue: https://github.com/copper-project/copper-rs/issues/120

## Detailed Changes

- Added culistid/CULISTID_UNKNOWN to CuLogEntry, appended to binary encoding, and defaulted on legacy decode in `core/cu29_log/src/lib.rs`; documented in `core/cu29_log/README.md`.
- Implemented culist context tracking (thread-local on std, atomic on no_std), guard helpers, and auto-population in log writes in `core/cu29_log_runtime/src/lib.rs`.
- Wrapped CopperList execution with culist guard in generated runtime code in `core/cu29_derive/src/lib.rs`.
- Propagated culist context into background threadpool tasks in `core/cu29_runtime/src/cuasynctask.rs`.
- Exposed culistid on Python bindings in `core/cu29_export/src/lib.rs`.
- Added tests for legacy decoding and culistid roundtrip in `core/cu29_log/src/lib.rs`, thread-local behavior in `core/cu29_log_runtime/src/lib.rs`, and an integration test plus config in `core/cu29_helpers/tests/culistid_logging.rs` and `core/cu29_helpers/tests/culistid_config.ron`. Added dev-dep on cu29 in `core/cu29_helpers/Cargo.toml`.

## Note
This PR serves as a prototype for discussion

### DIFF
--- a/core/cu29_log_runtime/Cargo.toml
+++ b/core/cu29_log_runtime/Cargo.toml
@@ -22,10 +22,18 @@ bincode = { workspace = true }
 smallvec = { workspace = true }
 spin = { workspace = true }
 log = { version = "0.4", default-features = false }
+portable-atomic = { version = "1.11", default-features = false, features = ["fallback"] }
 
 [dev-dependencies]
 cu29-value = { workspace = true }
 
 [features]
 default = ["std"]
-std = ["cu29-log/std", "cu29-traits/std", "cu29-clock/std", "bincode/std"]
+std = [
+    "cu29-log/std",
+    "cu29-traits/std",
+    "cu29-clock/std",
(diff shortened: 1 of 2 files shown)


makeecat:
> I think this is too costly for an information that is easy to get from the copperlists itself, let's give up on this. Thanks for checking that out!

Do you want to close the related issue and elaborate it there as well?

PR #666, opened 2026-01-17
ML inference example: yolo human pose for copper

gbin:


examples/cu_yolo_human_pose/src/tasks/gst_to_image.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        let image = CuImage::new(
+            CuImageBufferFormat {
+                width: self.width,
+                height: self.height,
+                stride: self.stride,
+                pixel_format: self.pixel_format,
+            },
+            CuHandle::new_detached(src.to_vec()),
```
gbin: This should be from a pool

PR #667, opened 2026-01-17
Adds CL stats to consolemon (Bandwidth stats)

gbin:
image

### DIFF
--- a/components/monitors/cu_consolemon/src/lib.rs
+++ b/components/monitors/cu_consolemon/src/lib.rs
@@ -46,11 +46,9 @@ use std::{collections::HashMap, io, thread};
 use tui_widgets::scrollview::{ScrollView, ScrollViewState};
 
 #[cfg(feature = "debug_pane")]
-const MENU_CONTENT: &str =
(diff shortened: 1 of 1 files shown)


components/monitors/cu_consolemon/src/lib.rs
```diff
@@ -856,6 +930,61 @@ impl UI {
         f.render_widget(table, area);
     }
 
+    fn draw_copperlist_stats(&self, f: &mut Frame, area: Rect) {
+        let stats = self.copperlist_stats.lock().unwrap();
+        let size_display = if stats.size_bytes > 0 {
+            format_bytes(stats.size_bytes as f64)
```
makeecat, line 936: The CL size should be self.size_byte * self.count?

My understanding:
If set_info receives CuListInfo { size_bytes: 65536, count: 2 } , the UI currently shows “CL size (RAM): 64 KiB”. The actual RAM footprint is 2 * 64 KiB = 128 KiB.

gbin: Ha I just wanted to know the size of one copperlist (ie. Cost of one cycle)

PR #669, opened 2026-01-18
(cu29_runtime) AlignedBuffer initializes to E::default() and guards zero-size/ZST cases

makeecat:
## Summary
AlignedBuffer initializes its storage to `E::default()` and guards zero-size cases, so `&[E]/&mut [E]` no longer expose uninitialized memory.
## Details
1. In `core/cu29_runtime/src/pool.rs`, `AlignedBuffer::new` clamps alignment to `align_of::<E>()`, initializes each element before returning
2. uses `layout.size()` in Drop to avoid deallocating zero-sized allocations
3. each unsafe block now has an explicit SAFETY rationale.

### DIFF
--- a/.config/_typos.toml
+++ b/.config/_typos.toml
@@ -7,6 +7,7 @@ IST = "IST"
 iterm = "iterm"
 NED = "NED"
 arange = "arange"
+mmaped = "mmaped"
 
 [files]
 extend-exclude = ["patches/*", "components/payloads/cu_ros_payloads/**"]
--- a/.github/workflows/reusable-unit-tests.yml
+++ b/.github/workflows/reusable-unit-tests.yml
@@ -62,6 +62,20 @@ jobs:
           log-file-suffix: '${{ matrix.os }}-${{ matrix.mode }}-${{ matrix.task }}-cuda.txt'
           use-github-cache: false
           use-local-cache: false
+      - name: Set CUDA compute capability (CI)
+        if: runner.os == 'Linux' && matrix.mode == 'cuda-debug'
+        shell: bash
+        run: |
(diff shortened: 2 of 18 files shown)


core/cu29_runtime/src/pool.rs
```diff
@@ -554,11 +555,28 @@ pub struct AlignedBuffer<E: ElementType> {
 
 impl<E: ElementType> AlignedBuffer<E> {
     pub fn new(num_elements: usize, alignment: usize) -> Self {
-        let layout = Layout::from_size_align(num_elements * size_of::<E>(), alignment).unwrap();
-        let ptr = unsafe { alloc(layout) as *mut E };
-        if ptr.is_null() {
+        let alignment = alignment.max(align_of::<E>());
```
gbin, line 562: I would simplify that... allocating 0 length buffer is probably a bug, I would just panic if 0 length maybe or return a CuError at least?

core/cu29_runtime/src/pool.rs
```diff
@@ -571,22 +589,23 @@ impl<E: ElementType> Deref for AlignedBuffer<E> {
...
     }
 }
 
 impl<E: ElementType> Drop for AlignedBuffer<E> {
     fn drop(&mut self) {
-        if !self.ptr.is_null() {
-            unsafe {
-                dealloc(self.ptr as *mut u8, self.layout);
-            }
+        if self.layout.size() != 0 {
```
gbin: then it should also be safe to remove the if

gbin (changes requested):

makeecat:
@codex address the comment from the requested changes in gbin's review

PR #681, opened 2026-01-20
fix(cu_caterpillar): correct justfile commands

TsaoLun:
Fixes incorrect params in `cu_caterpillar` justfile commands.
Related PR: #542 introduces justfile for this example to replace old bash script (not working anymore at that stage)

### DIFF
--- a/examples/cu_caterpillar/README.md
+++ b/examples/cu_caterpillar/README.md
@@ -7,8 +7,6 @@ See the crate cu29 for more information about the Copper project.
 
 ### Justfile commands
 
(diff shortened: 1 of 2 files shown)


makeecat:
@gbin lgtm for this PR.
However, `just dag` and `just mcap` is not working.
For `just dag`, we need to run `cargo run -p cu29-runtime --bin cu29-rendercfg -- --open copperconfig.ron` without `mission`, the justfile didn't process the empty argument of mission well.
For `just mcap`, the mcap export stuck forever

PR #689, opened 2026-01-20
[FEAT] composable SoA

makeecat:
## Summary

Add composable SoA with `#[soa(nested)]`, introducing storage-only types and nested-aware access/encode/serde paths, plus updated docs and tests for nested storage usage.

## Details

- `core/cu29_soa_derive/src/lib.rs`: parse `#[soa(nested)]`, generate `*SoaStorage`/`*SoaStorageWire`, route `access`/`apply`/`encode`/`serde` through nested storage, fix storage-aware Clone bounds, and expand inline docs (including ASCII layout).
- `core/cu29_soa_derive/tests/proctest.rs`: mark nested fields, update usage to nested storage accessors, and add nested serde + bincode round-trip coverage with length-mismatch validation.

## Notes

- Bincode layout changed: per-field length prefixes are removed; old blobs are not compatible.
- `#[soa(nested)]` does not support type aliases; use concrete types (and `::` absolute paths for external crates).

Continue the work of https://github.com/copper-project/copper-rs/tree/gbin/composable_soa

### DIFF
--- a/core/cu29_soa_derive/src/lib.rs
+++ b/core/cu29_soa_derive/src/lib.rs
@@ -79,6 +79,10 @@ use syn::{Attribute, Data, DeriveInput, Fields, Path, PathArguments, Type, parse
 /// ```
 /// For nested types from other crates, use absolute paths like `::other_crate::Type`
 /// so the generated storage path resolves correctly.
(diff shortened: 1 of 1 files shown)


gbin:
Do you mind adding what are the various memory layouts in each cases in the inline doc?

gbin (approved):

PR #736, opened 2026-01-26
Improved text logging architecture + colors

gbin:
Now the text logging is entirely the responsability of the monitor

NoMonitor just logs in plain text if textlogs is enabled + level CuLogMon same thing with structured logging colors CuConsoleMon same this but in the debug pane

image


examples/cu_caterpillar/Cargo.toml
makeecat, line 1: please leave a note in the PR explaining why it is needed. If it is needed, we may need to change template?

gbin: oh not at all sorry I was testing both monitors and I left that there... removed, thanks for spotting that

PR #744, opened 2026-01-27
feat: TimeTravelling Debugger API 

gbin:
Core / App API that works with the resim features to jump deterministically to anypoint of a log and replay steps with the possibility to inpects states and messages.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,36 @@
+name: CodeQL
+
+on:
+  schedule:
+    - cron: "0 3 * * 0"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  actions: read
+  security-events: write
+
+jobs:
+  analyze:
+    name: Analyze (Rust)
+    runs-on: ubuntu-latest
+
+    env:
+      WORKSPACE_SCOPE: workspace
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+            show-progress: false
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v4
+        with:
+          languages: rust
+          build-mode: none
+          queries: security-extended,security-and-quality
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v4
+        with:
+          category: "/language:rust"
--- a/.github/workflows/enforce_branch_name.yml
+++ b/.github/workflows/enforce_branch_name.yml
@@ -1,0 +1,37 @@
+name: PR Branch Name Check
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
+jobs:
+  check-source-branch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR source branch
+        run: |
+          if [[ "${{ github.event.pull_request.head.repo.fork }}" == "true" ]]; then
+            if [[ "${{ github.event.pull_request.head.ref }}" == "master" || "${{ github.event.pull_request.head.ref }}" == "main" ]]; then
+              echo "ERROR: Pull requests from the master/main branch of forks are not allowed, because it prevents maintainers from contributing to your PR."
+              echo "Please create a feature branch in your fork and submit the PR from that branch instead."
+              exit 1
+            fi
+          fi
+
+      - name: Leave comment if PR is from master/main branch of fork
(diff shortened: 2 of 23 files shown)


core/cu29_runtime/src/debug.rs
```diff
@@ -0,0 +1,605 @@
...
+            Err(e) => {
+                return Err(CuError::new_with_cause(
+                    "Failed to scan copperlist section",
+                    e,
+                ));
+            }
+        }
+    }
+    let first_id = first_id.ok_or_else(|| CuError::from("Empty copperlist section"))?;
+    let last_id = last_id.unwrap();
```
makeecat: `last_id.unwrap()` can panic if somehow `first_id` is set but `last_id` is not. The current logic should prevent that from happening. Would be nice to do a defensive coding here: maybe `unwrap_or(first_id)`?

core/cu29_runtime/src/debug.rs
```diff
@@ -0,0 +1,605 @@
...
+            total_entries,
+            keyframes,
+            started: false,
+            current_idx: None,
+            last_keyframe: None,
+            build_callback,
+            time_of,
+            cache: HashMap::new(),
+            cache_order: VecDeque::new(),
+            cache_cap: 8,
```
makeecat: better handeling magic number?

core/cu29_runtime/src/copperlist.rs
```diff
@@ -49,7 +49,7 @@ impl Display for CopperListState {
     }
 }
 
-#[derive(Debug, Encode, Decode, Serialize, Deserialize)]
+#[derive(Debug, Encode, Decode, Serialize, Deserialize, Clone)]
```
makeecat: Explain in the PR description why Clone is needed

core/cu29_runtime/src/debug.rs
```diff
@@ -0,0 +1,605 @@
...
+                    std::cmp::Ordering::Less
+                } else {
+                    std::cmp::Ordering::Equal
+                }
+            })
+            .ok()
+    }
+
+    fn find_section_for_time(&self, ts: CuTime) -> Option<usize> {
+        self.sections
```
makeecat: change to `binary_search_by` to be consistent?

core/cu29_runtime/src/debug.rs
```diff
@@ -0,0 +1,605 @@
...
+    let last_id = last_id.unwrap();
+    Ok((count, first_id, last_id, first_ts, last_ts))
+}
+
+/// Read a specific section at a given position from disk.
+fn read_section_at(log_base: &Path, pos: LogPosition) -> CuResult<(SectionHeader, Vec<u8>)> {
+    let logger = UnifiedLoggerBuilder::new()
+        .file_base_name(log_base)
+        .build()
+        .map_err(|e| CuError::new_with_cause("Failed to open unified log", e))?;
```
makeecat, line 633: Should we hold a `UnifiedLoggerRead` handle in the session struct?

core/cu29_runtime/src/debug.rs
```diff
@@ -0,0 +1,605 @@
...
+                } else if culistid > s.last_id {
+                    std::cmp::Ordering::Less
+                } else {
+                    std::cmp::Ordering::Equal
+                }
+            })
+            .ok()
+    }
+
+    fn find_section_for_time(&self, ts: CuTime) -> Option<usize> {
```
makeecat, line 268: Add a definition of this function, some users may assume it is a lower-bound (lower-bound defined as: the first section whose `first_ts >= ts` or if `ts` is earlier than the first section, pick that first section, only error when `ts` is beyond the last section's range)

core/cu29_runtime/src/debug.rs
```diff
@@ -0,0 +1,605 @@
...
+                    }
+                }
+            }
+            _ => {
+                // ignore other sections
+            }
+        }
+    }
+
+    // Optional down-sampling of index to reduce memory for huge logs
```
makeecat: This part of the code returns `total_entries` which was computed from the original sections. After filtering, the `start_idx` values in the filtered sections still reference the original indices, but lookups using
  `find_section_for_index()` will fail because binary search won't find matching sections for indices that were removed. Also, `find_section_for_index()` uses binary search on `start_idx`/`len` to locate sections. If sections are filtered out, there will be gaps in coverage

core/cu29_runtime/src/debug.rs
gbin: yeah this is super wasteful

PR #760, opened 2026-01-28
chore: Add cu-serial-util and refactor CRSF/MSP bridge batching

makeecat:
## Summary

- Introduce `cu-serial-util` (no_std) with shared LockedSerial
- Switch `CRSF/MSP` bridges to use shared serial wrapper and drop `spin` deps
- Refactor `MSP` request/response batching into generic `MspBatch`
- Update `cu-elrs-bdshot-demo` to use cu-serial-util`

## Details

- CRSF/MSP bridges: remove local `LockedSerial`, use `cu_serial_util::LockedSerial`
- MSP bridge: unify batch encoding/decoding and simplify length checks
- `cu_elrs_bdshot_demo`: adds dependency and updates serial type alias

### DIFF
--- a/.github/workflows/first_time_contrib.yml
+++ b/.github/workflows/first_time_contrib.yml
@@ -15,13 +15,12 @@ jobs:
     steps:
       - uses: actions/first-interaction@v3.1.0
         with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-          issue-message: |
+          issue_message: |
             Hi! Thanks for opening this issue.
 
             Because this is your first time contributing to this repository, please read our contributor guide:
             https://github.com/copper-project/copper-rs/blob/master/CONTRIBUTING.md
-          pr-message: |
+          pr_message: |
             Hi! Thanks for opening this pull request.
 
             Because this is your first time contributing to this repository, please read our contributor guide:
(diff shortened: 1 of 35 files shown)


components/bridges/cu_crsf/src/lib.rs
```diff
@@ -273,41 +271,11 @@ impl ResourceBundle for StdSerialBundle {
...
-impl<T: Write> Write for LockedSerial<T> {
-    fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
-        let mut guard = self.0.lock();
-        guard.write(buf)
-    }
-
-    fn flush(&mut self) -> Result<(), Self::Error> {
-        let mut guard = self.0.lock();
-        guard.flush()
+        manager.add_owned(key, cu_serial_util::LockedSerial::new(serial))
```
gbin, line 274: Big picture here, we are doing something wrong somewhere.

Serial ports? they are NEVER shared between bridges.
They are exclusive resources. See add_owned :). Actually this resource manager should not even be there, this is a temporary thing. 

So having a spin::Mutex on them, I think we iteratively went in the wrong direction.

They should be owned and taken from the resources then exclusively used by their respective bridges.

If we have some forced useless Sync due to some ownership / API constraints -> maybe we can unsafe our way out of this?

Let's discuss this on discord.

gbin (changes requested):
see comment, I think we need a larger refactor here. Probably a resource manager for std platforms for various resources like serial ports, a little bit like we did for the flight controller but for a normal linux. Let's discuss on discord

PR #763, opened 2026-01-28
chore(cu29_helpers): simplify basic_copper_setup

makeecat:
remove redundant `let extra: Option<TermLogger> = None;`

### DIFF
--- a/core/cu29_helpers/src/lib.rs
+++ b/core/cu29_helpers/src/lib.rs
@@ -54,10 +54,6 @@ pub fn basic_copper_setup(
         4096 * 10,
     )?;
 
-    #[cfg(debug_assertions)]
-    let extra: Option<TermLogger> = None;
-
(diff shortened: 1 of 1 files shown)


core/cu29_helpers/src/lib.rs
```diff
@@ -54,10 +54,6 @@ pub fn basic_copper_setup(
         4096 * 10,
     )?;
 
-    #[cfg(debug_assertions)]
-    let extra: Option<TermLogger> = None;
-
-    #[cfg(not(debug_assertions))]
     let extra: Option<TermLogger> = None;
```
gbin, line 58: TermLogger is not gone completely actually?

makeecat: I assume we still need one line? I am not sure if we can completely remove it

PR #768, opened 2026-01-29
feat: Add Rerun-based log visualization support via a new cu29-logviz

### LINKED ISSUE #695: Implement logviz with rerun
Today people using rerun including myself just plug the rerun 1 liner in a task, run the robot live and debug it with it.

It is cool to correlate directly what is happening live with the robot and debug.

But another way of using it could be to automatically generate the right rerun types (or simply implement them in the standard payloads!) and straight visualize a copper log in rerun. 

Note: I tried it through Mcap but while foxglove can read what we generate, rerun cannot and I think the level of integration will be much better if we just call the rerun API directly (this API is a work of art and it is much more expressive to hint towards the UI representation than simple data dumps)

makeecat:
## Summary

- Add Rerun-based log visualization support via a new `cu29-logviz` crate and generated `logviz` emitters.
- Introduce a `cu-logviz-demo` example with standard and custom `logviz` binaries plus updated `README` guidance.
## Related issues
- Closes #695 

## Details

- `core/cu29_logviz` provides log helpers for images, point clouds, transforms, IMU, time-of- validity handling, and a JSON fallback logger, plus as_components adapters for Rerun component logging.
- `gen_cumsgs!` now emits a `LogvizDataSet impl` behind `cfg(feature = "logviz")`, with type-based dispatch for `CuImage<Vec<u8>>`, `PointCloud`, `PointCloudSoa`, `Transform3D<f32/f64>`, and `muPayload`, falling back to JSON flattening for other payloads.
- New example examples/cu_logviz_demo generates logs and exposes `-logviz` and `-logviz-custom` binaries; includes sample tasks, config, and tests.
- Workspace updates: add `cu29-logviz` and `cu-logviz-demo`, add `cu-spatial-payloads` and `serde_json`, and refresh compile-fail snapshots for the new `logviz` cfg.

## Impact
- Users can visualize Copper logs in Rerun via an app-specific logviz binary generated by `gen_cumsgs!`.
- Adds an optional `logviz` feature path with new dependencies and test coverage, plus updated documentation for usage.
    
### Test
```
# Generate a log
cargo run -p cu-logviz-demo --bin cu-logviz-demo
# Visualize it in Rerun
cargo run -p cu-logviz-demo --bin cu-logviz-demo-logviz --features logviz -- logs/logviz_demo.copper --spawn

# Visualize it in Rerun with the custom frame tree (map/base_link/...)
cargo run -p cu-logviz-demo --bin cu-logviz-demo-logviz-custom --features logviz -- logs/logviz_demo.copper --spawn
```

### DIFF
--- a/components/payloads/cu_spatial_payloads/Cargo.toml
+++ b/components/payloads/cu_spatial_payloads/Cargo.toml
@@ -12,6 +12,7 @@ repository.workspace = true
 
 [dependencies]
 bincode = { workspace = true }
+cu29 = { path = "../../../core/cu29", version = "0.12.0", default-features = false }
 uom = { workspace = true }
 serde = { workspace = true }
 rerun = { workspace = true, optional = true }
--- a/components/payloads/cu_spatial_payloads/src/lib.rs
+++ b/components/payloads/cu_spatial_payloads/src/lib.rs
@@ -1,6 +1,7 @@
 use bincode::{Decode, Encode};
 use core::fmt::Debug;
 use core::ops::Mul;
+use cu29::prelude::*;
 use serde::{Deserialize, Serialize};
 use uom::si::angle::radian;
 use uom::si::f32::Angle as Angle32;
@@ -17,7 +18,8 @@ mod rerun_components;
 
 /// Transform3D represents a 3D transformation (rotation + translation)
 /// When the glam feature is enabled, it uses glam's optimized types internally
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Reflect)]
+#[reflect(opaque, from_reflect = false)]
 pub struct Transform3D<T: Copy + Debug + 'static> {
     #[cfg(feature = "glam")]
     inner: TransformInner<T>,
--- a/core/cu29_soa_derive/src/lib.rs
+++ b/core/cu29_soa_derive/src/lib.rs
@@ -130,6 +130,25 @@ pub fn derive_soa(input: TokenStream) -> TokenStream {
 
     let input = parse_macro_input!(input as DeriveInput);
     let visibility = &input.vis;
+    let derive_reflect = input
+        .attrs
+        .iter()
+        .any(|attr| attr.path().is_ident("reflect"));
+    let reflect_import = if derive_reflect {
+        quote!(
+            use super::{bevy_reflect, Reflect};
+        )
+    } else {
+        quote!()
+    };
+    let soa_reflect_attrs = if derive_reflect {
+        quote! {
+            #[derive(Reflect)]
+            #[reflect(from_reflect = false)]
+        }
+    } else {
+        quote!()
+    };
 
     let name = &input.ident;
     let module_name = format_ident!("{}_soa", name.to_string().to_lowercase());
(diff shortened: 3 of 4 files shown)


core/cu29_derive/src/lib.rs
```diff
@@ -50,6 +50,149 @@ fn rtsan_guard_tokens() -> proc_macro2::TokenStream {
     }
 }
 
+fn type_last_ident(ty: &Type) -> Option<String> {
```
gbin: Do we need logviz to be computed at compile time? 
For runtime stuff, I would say yes but here it is offline, for the sake of complexity we cannot just have the type mapping done in the standard payload crates as traits or implementation of traits to get their rerun types?

makeecat: I was thinking maybe we could implement streaming with rerun as well in the future if we support it at compile time?


If our scope is supporting logviz only at offline time, yes, I agree with you.

Which way should we go?

gbin: ha, I see where you are coming from ... hmm. The problem of that approach is that it is not generic nor scalable: you need to change Copper the compiler every time you add a standard message ... then what about the users one? they need to fork Copper?

This is making a soft, stringly typed, reverse dependency: core <- payloads <- core

I think it is self evident that specific msg datatypes have nothing to do in the code generation path. If we want to accelerate the R/T viz we can add helper macros maybe. But imagine, this shit will hit apache arrow or whatever it is, at that point, saving 3ns by not calling a interface insta becomes negligeable.

So let's do dynamic traits as rerun intend for our own types for now, if someday this becomes a bottleneck we will make some macro from a common payload crate.

core/cu29_logviz/src/as_components.rs
```diff
@@ -0,0 +1,60 @@
...
+    inner: &'a CuImage<Vec<u8>>,
+}
+
+impl<'a> LogvizImageVec<'a> {
+    pub fn new(inner: &'a CuImage<Vec<u8>>) -> Self {
+        Self { inner }
+    }
+}
+
+impl<'a> AsComponents for LogvizImageVec<'a> {
```
gbin: I would push that to all the standard types in Copper and enable the AsComponents under a rerun flag over there.

Like that we have if for all the 3 tools we want:
1. live viz
2. log viz
3. debugger

makeecat: Applied in the payload layer with feature flags:
- added `rerun` features in standard payload crates used here
- moved `AsComponents` impls to payload crates
- updated `cu29-logviz` and the demo to consume payload traits directly

This gives shared type-to-rerun mapping that can be reused by log viz today and by live viz/debugger paths without duplicating conversion logic in `cu29-logviz`.

makeecat:
@codex review

components/payloads/cu_spatial_payloads/src/rerun_components.rs
```diff
@@ -0,0 +1,93 @@
...
+use crate::Transform3D;
+
+fn transform_parts<T: Copy + Into<f64> + Debug + Default + 'static>(
+    transform: &Transform3D<T>,
+) -> ([f32; 3], [[f32; 3]; 3]) {
+    let matrix = (*transform).to_matrix();
+    let translation = [
+        matrix[3][0].into() as f32,
+        matrix[3][1].into() as f32,
+        matrix[3][2].into() as f32,
```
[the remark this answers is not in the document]
makeecat: @codex How should we fix it?

PR #790, opened 2026-02-07
feat: add cu-linux-resources and igrates Linux/std components to consume centralized serial/I2C/GPIO resources

makeecat:
## Summary
This PR introduces a new shared resource crate, `cu-linux-resources`, and migrates several Linux/std components to consume centralized serial/I2C/GPIO resources instead of opening devices ad hoc.

## Changes
  - Added new workspace crate: `components/res/cu_linux_resources` (`cu-linux-resources` package, `cu_linux_resources` module path).
  - New `LinuxResources` bundle can register:
      - serial via `LinuxSerialPort`
      - `i2c` via `LinuxI2c`
      - `gpio_out` via `LinuxOutputPin`
      - `gpio_in` via `LinuxInputPin`
  - Supports config aliases/fallbacks:
      - Serial: serial_path or device, with baudrate and timeout_ms
      - I2C: i2c_bus or bus
      - GPIO out: gpio_out_pin or pin
      - GPIO in: gpio_in_pin
  - Migrated components to use these resource types and optional resource bindings:
      - Bridges: `cu_crsf`, `cu_msp_bridge`
      - Sinks: `cu_lewansoul`, `cu_rp_gpio`
      - Sources: `cu_rp_encoder`, `cu_wt901`
  - Updated `docs/examples` to show provider: `cu_linux_resources::LinuxResources` and resource wiring (e.g. `<bundle>.serial`, `<bundle>.gpio_in`).
  - `StdSerialBundle` compatibility remains in CRSF/MSP bridges, now backed by LinuxSerialPort.
## Testing
- [x] `just std-ci`
- [x] `just lint`
- [x] `cargo +stable nextest run --workspace --all-targets`
- [x] `just nostd-ci`

## Checklist
- [x] I have added or updated tests where needed
- [x] I have considered platform impact (Linux/macOS/Windows/embedded)
- [x] I have considered config/logging changes (if applicable)
- [x] This change is not a breaking change (or I documented it below)

## Breaking changes (if any)
  - Custom resource providers/bindings: resource concrete types changed to cu_linux_resources wrappers (LinuxSerialPort, LinuxI2c, LinuxOutputPin, LinuxInputPin), so type-matching bindings may need updates.
  - if migrating CRSF from legacy StdSerialBundle to LinuxResources without explicit baudrate, default may differ (420000 legacy vs 115200 LinuxResources).
## Additional context
N/A


components/bridges/cu_crsf/src/lib.rs
```diff
@@ -223,19 +210,39 @@ where
...
-    use std::time::Duration;
-
-    pub type StdSerial = FromStd<Box<dyn serialport::SerialPort>>;
-
-    pub fn open(path: &str, baud: u32, timeout_ms: u64) -> std::io::Result<StdSerial> {
-        let port = serialport::new(path, baud)
-            .timeout(Duration::from_millis(timeout_ms))
-            .open()?;
-        Ok(FromStd::new(port))
+pub struct LockedSerial<T>(pub T);
```
gbin: I don't understand why the serial users have to implement anything after we move everything to the bundles, especially LockedSerial vs Owned serial port. It should become uniform with the embedded traits as a common denominator across std and nostd

components/bridges/cu_crsf/src/lib.rs
```diff
@@ -258,56 +265,26 @@ impl ResourceBundle for StdSerialBundle {
...
-        let serial = std_serial::open(&path, baud, timeout_ms).map_err(|err| {
-            CuError::from(format!(
-                "Failed to open serial `{path}` at {baud} baud: {err}"
-            ))
-        })?;
+        let baud = cfg.get::<u32>("baudrate")?.unwrap_or(420_000);
+        let timeout_ms = cfg.get::<u64>("timeout_ms")?.unwrap_or(100);
+
+        let serial =
+            cu_linux_resources::LinuxSerialPort::open(&path, baud, timeout_ms).map_err(|err| {
```
gbin: here too I don't get it, it should be handled on the bundle side not the user code:

you set up your port parameters in the bundle and it should be able to open it and give it to the task without any manual tweeking.

components/bridges/cu_msp_bridge/src/lib.rs
```diff
@@ -365,21 +357,39 @@ where
...
-    pub type StdSerial = FromStd<Box<dyn serialport::SerialPort>>;
-
-    pub fn open(path: &str, baud: u32, timeout_ms: u64) -> std::io::Result<StdSerial> {
-        let port = serialport::new(path, baud)
-            .timeout(Duration::from_millis(timeout_ms))
-            .open()?;
-        Ok(FromStd::new(port))
+pub struct LockedSerial<T>(pub T);
+
+impl<T> LockedSerial<T> {
```
gbin: same

components/bridges/cu_msp_bridge/src/lib.rs
```diff
@@ -403,53 +413,23 @@ impl ResourceBundle for StdSerialBundle {
             ))
         })?;
         let device = cfg
```
gbin: same

components/res/cu_linux_resources/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+pub const GPIO_PIN_KEY: &str = "pin";
+
+#[cfg(feature = "std")]
+pub const DEFAULT_SERIAL_BAUDRATE: u32 = 115_200;
+#[cfg(feature = "std")]
+pub const DEFAULT_SERIAL_TIMEOUT_MS: u64 = 50;
+
+/// Wrapper for resources that are logically exclusive/owned by a single
+/// component but still need to satisfy `Sync` bounds at registration time.
+pub struct Exclusive<T>(T);
```
gbin, line 67: this is basically what the users of the bundle should be using? Why do we need to wrap it?

components/bridges/cu_crsf/src/lib.rs
makeecat: Fixed. removed LockedSerial from both bridge crates and the bridges now consume `Owned<S>` directly. Serial adaptation is no longer in bridge usercode.

components/bridges/cu_crsf/src/lib.rs
makeecat: Fixed on the bundle side. Serial open/config stays in resource bundles (`LinuxResources / legacy StdSerialBundle`) with config keys (`serial_path/device, baudrate, timeout_ms`), and the bridge receives the ready resource via mapping

components/bridges/cu_msp_bridge/src/lib.rs
makeecat: Applied same change for `MSP`: removed bridge-local serial wrapper and kept resource acquisition as `Owned<S>` from bundle mapping

components/bridges/cu_msp_bridge/src/lib.rs
makeecat: serial setup remains in bundle config/provider, and bridge user path is now uniform without extra adapter code

components/res/cu_linux_resources/src/lib.rs
makeecat: Yes, this is the bundle-level wrapper users should use when needed. We still need `Exclusive<T>` at registration because `ResourceManager::add_owned `requires T: `Send + Sync`. I kept that adaptation centralized in `cu_linux_resources`, added `embedded-io 0.7` forwarding, added a test

makeecat:
> It is going in the right direction but it looks half done. The users of the serial ports should just get a serial port, no adapter needed from both std and nostd implementations. (but if I missed some hard requirement please comment)

Serial consumer-side adapters were removed; serial adaptation now lives in bundle/resource layer only, with `no-std` migration to `cu_linux_resources::Exclusive<T>` and validation across affected crates/examples.

components/bridges/cu_crsf/src/lib.rs
gbin: I still don't understand, we should have no legacy stuff, everything in either a board bundle or a linux bundle at that stage. If we miss a bundle for the rp2350 we can add one too.

components/bridges/cu_msp_bridge/src/lib.rs
```diff
@@ -403,53 +377,23 @@ impl ResourceBundle for StdSerialBundle {
...
-            .get::<String>(DEVICE_KEY)?
+            .get::<String>("device")?
             .filter(|s| !s.is_empty())
-            .unwrap_or_else(|| DEFAULT_DEVICE.to_string());
-        let baudrate = cfg.get::<u32>(BAUD_KEY)?.unwrap_or(DEFAULT_BAUDRATE);
-        let timeout_ms = cfg.get::<u64>(TIMEOUT_KEY)?.unwrap_or(DEFAULT_TIMEOUT_MS);
+            .unwrap_or_else(|| "/dev/ttyUSB0".to_string());
+        let baudrate = cfg.get::<u32>("baudrate")?.unwrap_or(115_200);
```
gbin: those configs should be in bundles

components/bridges/cu_msp_bridge/src/lib.rs
```diff
@@ -403,53 +377,23 @@ impl ResourceBundle for StdSerialBundle {
...
+        let timeout_ms = cfg.get::<u64>("timeout_ms")?.unwrap_or(50);
 
-        let serial = std_serial::open(&device, baudrate, timeout_ms).map_err(|err| {
-            CuError::from(format!(
-                "MSP bridge failed to open serial `{device}` at {baudrate} baud: {err}"
-            ))
-        })?;
+        let serial = cu_linux_resources::LinuxSerialPort::open(&device, baudrate, timeout_ms)
```
gbin: that should happen in the bundle

components/sinks/cu_lewansoul/src/lib.rs
```diff
@@ -72,11 +74,40 @@ fn angle_to_position(angle: Angle) -> i16 {
...
     ids: [u8; 8], // TODO: WIP
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Binding {
+    Serial,
+}
+
+#[derive(Default)]
+pub struct LewansoulResources {
```
gbin, line 85: this should be a standard bundle

components/sinks/cu_lewansoul/src/lib.rs
```diff
@@ -229,14 +252,28 @@ impl CuSinkTask for Lewansoul {
...
+            None => {
+                let serial_dev: String = kv
+                    .get("serial_dev")
+                    .ok_or(
+                        "Lewansoul expects a serial_dev config entry or a `serial` resource binding.",
+                    )?
+                    .clone()
+                    .into();
+
+                let port = serialport::new(serial_dev.as_str(), SERIAL_SPEED)
```
gbin: we should always build the serial port in the bundle.

components/sinks/cu_rp_gpio/src/lib.rs
```diff
@@ -17,12 +16,51 @@ fn gpio() -> &'static Gpio {
     GPIO.get_or_init(|| Gpio::new().expect("Could not create GPIO bindings"))
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Binding {
+    Pin,
+}
+
+#[derive(Default)]
+pub struct GpioResources {
```
gbin, line 16: this should be taken from the linux bundle

components/sinks/cu_rp_gpio/src/lib.rs
```diff
@@ -94,3 +133,16 @@ impl CuSinkTask for RPGpio {
         Ok(())
     }
 }
+
+fn pin_from_config(config: Option<&ComponentConfig>) -> CuResult<u8> {
```
gbin, line 124: the config should be in the bundle (look at LED for the STM32 this is the same thing)

components/sources/cu_rp_encoder/src/lib.rs
```diff
@@ -4,17 +4,75 @@ mod mock;
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Binding {
+    ClkPin,
+    DatPin,
+}
+
+#[derive(Default)]
+pub struct EncoderResources {
+    #[cfg(hardware)]
+    pub clk_pin: Option<Owned<InputPin>>,
```
gbin: those should be configured and available from the bundle

components/sources/cu_wt901/src/lib.rs
```diff
@@ -62,7 +68,46 @@ use uom::fmt::DisplayStyle::Abbreviation;
...
+    i2c: LinuxI2c,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Binding {
+    I2c,
+}
+
+#[derive(Default)]
+pub struct Wt901Resources {
```
gbin, line 77: again those I2c should not be floating here at all... If I want to use that with an i2c from stm32? This is the point of the resourcebundles that we don't make one PER task, it should represent a platform and all the tasks can use them

gbin (changes requested):
ok I think the LLM doesn't understand at all what we are trying to do here.

components/bridges/cu_crsf/src/lib.rs
makeecat: addressed in  c7623da

components/bridges/cu_msp_bridge/src/lib.rs
makeecat: addressed in cb7df73

components/bridges/cu_msp_bridge/src/lib.rs
makeecat: adressed in 5a6088e

components/sinks/cu_lewansoul/src/lib.rs
makeecat: addressed in 60553b2

components/sinks/cu_lewansoul/src/lib.rs
makeecat: addressed in 4c28087

components/sinks/cu_rp_gpio/src/lib.rs
makeecat: addressed in a83c962

components/sinks/cu_rp_gpio/src/lib.rs
makeecat: addressed in 5047a61

components/sources/cu_rp_encoder/src/lib.rs
makeecat: adressed in ac298ae

components/sources/cu_wt901/src/lib.rs
makeecat: addressed in e4bf94e

components/res/cu_linux_resources/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+    fn is_high(&mut self) -> Result<bool, Self::Error> {
+        self.0.is_high()
+    }
+
+    fn is_low(&mut self) -> Result<bool, Self::Error> {
+        self.0.is_low()
+    }
+}
+
+#[cfg(feature = "std")]
```
gbin: I think it is better to just say this is an std component so we can remove all the conditionals.

components/bridges/cu_msp_bridge/README.md
```diff
@@ -16,7 +18,7 @@ The bridge expects a `serial` resource. For std targets use the built-in `StdSer
 resources: [
   (
     id: "fc",
-    provider: "cu_msp_bridge::StdSerialBundle",
+    provider: "cu_linux_resources::LinuxResources",
```
gbin, line 25: This is still not following the pattern we have on embedded. For example:
we declare a bundle on the resource part:

```Rust
// Resource bundle for the MicoAir H743 board.
pub struct MicoAirH743;

bundle_resources!(
    MicoAirH743: Uart6, Uart2, Uart4, GreenLed, Logger, Bmi088Spi, Bmi088AccCs, Bmi088GyrCs, Bmi088Delay, BatteryAdc
);

```

Then we pick one, and use it:

```Rust
use super::*;
resources!({
    led => Owned<spin::Mutex<GreenLed>>,
});

#[derive(Reflect)]
#[reflect(from_reflect = false)]
pub struct ActivityLed {
    on: bool,
    #[reflect(ignore)]
    led: spin::Mutex<GreenLed>,
}

impl Freezable for ActivityLed {}

impl CuSinkTask for ActivityLed {
    type Input<'m> = CuMsg<ControlInputs>;
    type Resources<'r> = Resources;
```

The only difference is that we have some hardcoding to do.

maybe something like that:

pub struct LinuxResources;

bundle_resources!(
    LinuxResources: USBUart0, USBUart1, USBUart2, I2C0, I2C1 ... 
);

or maybe some more tricky macro to add some flexibility?

gbin:
I  would just expose a set of hardcoded peripherals in the bundle for now:

like /dev/ttyACM0 /dev/ttyACM1 /dev/ttyACM2 then /dev/ttyUSB0 /dev/ttyUSB1 /dev/ttyUSB2 then /dev/i2c-0 /dev/i2c-1 /dev/i2c-2 ... 

you can imagine those as "slots" the tasks can take

in just dag it should exactly look like this  but with a bunch of unused resources
image

components/bridges/cu_msp_bridge/README.md
makeecat: Implemented in `79e52e1fe`:
- `pub struct LinuxResources;`
- `bundle_resources!`

bundle now follows the embedded pattern: fixed resource IDs declared once in the resource crate, then consumers pick slots via normal resource bindings (for example `fc.serial_usb0`).

The hardcoded Linux defaults are centralized in the bundle.

Updated DAG cataloging so `just dag` shows the full slot inventory, including unused slots, matching your expected view.

components/res/cu_linux_resources/src/lib.rs
makeecat: removed the redundant non-Linux no-op block in `a69fb28`

components/bridges/cu_msp_bridge/README.md
```diff
@@ -4,30 +4,32 @@ Bridge that combines the MSP source/sink tasks into one serial transport.  It ex
...
 
-| Key          | Type   | Default        | Description                                                |
-|--------------|--------|----------------|------------------------------------------------------------|
-| `device`     | string | `/dev/ttyUSB0` | Path to the serial device (can be a PTY).                  |
-| `baudrate`   | u32    | `115200`       | Serial baud rate.                                          |
-| `timeout_ms` | u64    | `50`           | Read timeout passed to the serial driver in milliseconds.  |
+| Key                | Type   | Default        | Description                                                     |
+|--------------------|--------|----------------|-----------------------------------------------------------------|
+| `tty_usb0_path`    | string | `/dev/ttyUSB0` | Path for the `serial_usb0` slot (can be a PTY).                |
+| `serial_baudrate`  | u32    | `115200`       | Serial baud rate for all Linux serial slots.                   |
```
gbin: this is not possible.. you cannot set the baudrate for ALL the serial ports at the same speed.

check how we do this per uart in the mico bundle.

If we allow a path per serial then anonymize them:

serial0_path
serial0_baudrate
serial0_parity
serial0_stopbits
serial1_path
...

then the user can assign a USB or ACM path and you are good to go.

components/bridges/cu_msp_bridge/README.md
```diff
@@ -4,30 +4,32 @@ Bridge that combines the MSP source/sink tasks into one serial transport.  It ex
...
+| `serial_timeout_ms`| u64    | `50`           | Read timeout passed to the serial driver in milliseconds.      |
 
 ```ron
 resources: [
   (
     id: "fc",
```
gbin: nit: id it with something else than fc :)

components/res/cu_linux_resources/src/lib.rs
```diff
@@ -0,0 +1,650 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
gbin: IMHO don't provide no_std because if this is linux, this is std...  just the filtering here is ridiculous, basically the whole file.

components/bridges/cu_msp_bridge/README.md
makeecat: addressed in 6575024

PR #809, opened 2026-02-12
fix:  default LinuxResources points to non existent devices

### LINKED ISSUE #805: bug: default LinuxResources points to non existent devices
### Summary

all the resources try to initialize on existent or non existent devices

### Steps to reproduce

Run any example based on serial ports and by default all the resources try to initialize on non existent devices:

LinuxResources: skipping serial slot serial_acm1 (dev /dev/ttyACM1): No such file or directory
LinuxResources: skipping serial slot serial_acm2 (dev /dev/ttyACM2): No such file or directory
LinuxResources: skipping serial slot serial_usb0 (dev /dev/ttyUSB0): No such file or directory
LinuxResources: skipping serial slot serial_usb1 (dev /dev/ttyUSB1): No such file or directory
LinuxResources: skipping serial slot serial_usb2 (dev /dev/ttyUSB2): No such file or directory

### Expected behavior

I must have missed this in review.

So
1. we were supposed to switch to serial0 serial1 etc... no more serial_acm1
2. we should NOT affect default stuff to them trying to blindly initialize peripheral EVEN if we don't need them
3. we should NOT eprintln -> use the Copper logger.

### Actual behavior

It prints on stderr, initialize peripheral randomly (if the user wants to actually use them they need to put random strings in there).

### Logs / backtrace

```shell

```

### Operating system

macOS

### Rust toolchain

whatever

### Copper revision

_No response_

### Regression

- [ ] This is a regression

### Additional context

_No response_

makeecat:
## Summary
fix the bug that default LinuxResources points to non existent devices introduced by PR: #790 
## Related issues
- Closes #805 

## Changes
- Renamed slot identifiers: SerialAcm0..SerialUsb2 → Serial0..Serial5
- Removed default device paths: Serial and I2C slots are only initialized when explicitly configured via serialN_dev or i2cN_dev
- Replaced eprintln! with warning!: Using Copper's unified logging system (required adding build.rs with LOG_INDEX_DIR setup and
converting errors to strings via .to_string())
- Updated documentation: README files for cu_linux_resources, cu_crsf, cu_msp_bridge, cu_lewansoul, and rendercfg.rs
- Updated example config: cu_msp_bridge_loopback now uses linux.serial3 instead of linux.serial_usb0
## Testing
- [x] `just std-ci`
- [x] `just lint`
- [x] `cargo +stable nextest run --workspace --all-targets`

## Checklist
- [x] I have added or updated tests where needed
- [x] I have considered platform impact (Linux/macOS/Windows/embedded)
- [x] I have considered config/logging changes (if applicable)
- [x] This change is not a breaking change (or I documented it below)

### DIFF
--- a/components/bridges/cu_crsf/README.md
+++ b/components/bridges/cu_crsf/README.md
@@ -14,21 +14,13 @@ The bridge expects a `serial` resource (anything implementing `embedded_io` `Rea
 
 Use `cu_linux_resources::LinuxResources` as the serial provider; it stores owned
 resources in fixed slots like `<bundle>.serial0` through `<bundle>.serial5`.
-
-For example, to use `serial3`, configure serial slot `serial3_*`:
-
-| Key | Type | Default | Description |
-| --- | --- | --- | --- |
-| `serial3_dev` | string | (required) | Device node for the `serial3` resource slot. |
-| `serial3_baudrate` | u32 | `115200` | UART baudrate for this slot. |
-| `serial3_parity` | string | `none` | Parity for this slot (`none`, `odd`, `even`). |
-| `serial3_stopbits` | u8 | `1` | Stop bits for this slot (`1` or `2`). |
-| `serial3_timeout_ms` | u64 | `50` | Read timeout for this slot in milliseconds. |
+For serial config keys (`serialN_dev`, `serialN_baudrate`, parity/stopbits/timeout), see
+[`cu_linux_resources` README: Config Keys / Serial](../../res/cu_linux_resources/README.md#serial).
 
 ```ron
 resources: [
   (
-    id: "radio",
+    id: "linux",
     provider: "cu_linux_resources::LinuxResources",
     config: { "serial3_dev": "/dev/ttyUSB0", "serial3_baudrate": 420000 },
   ),
@@ -37,7 +29,7 @@ bridges: [
   (
     id: "crsf",
     type: "cu_crsf::CrsfBridgeStd",
-    resources: { serial: "radio.serial3" },
+    resources: { serial: "linux.serial3" },
     channels: [ Rx (id: "rc_rx"), Rx (id: "lq_rx") ],
   ),
 ],
--- a/components/bridges/cu_msp_bridge/README.md
+++ b/components/bridges/cu_msp_bridge/README.md
@@ -7,16 +7,8 @@ Bridge that combines the MSP source/sink tasks into one serial transport. It exp
 The bridge expects a `serial` resource. For std targets, use
 `cu_linux_resources::LinuxResources` as the provider. It opens serial devices
 and exposes fixed serial slots such as `<bundle>.serial0` through `<bundle>.serial5`.
-
-For example, to use `serial3`, configure serial slot `serial3_*` in the Linux bundle:
-
-| Key | Type | Default | Description |
-| --- | --- | --- | --- |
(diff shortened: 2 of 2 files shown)


components/res/cu_linux_resources/src/lib.rs
```diff
@@ -635,7 +611,7 @@ fn parse_serial_stop_bits_value(raw: u8) -> CuResult<SerialStopBits> {
 }
 
 fn slot_name(id: LinuxResourcesId) -> &'static str {
-    LINUX_RESOURCE_SLOT_NAMES[id as usize]
+    LINUX_RESOURCES_ID_NAMES[id as usize]
```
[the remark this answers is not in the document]
makeecat: fixed in 14390a7

components/bridges/cu_crsf/README.md
```diff
@@ -37,7 +37,7 @@ bridges: [
   (
     id: "crsf",
     type: "cu_crsf::CrsfBridgeStd",
-    resources: { serial: "radio.serial_usb0" },
+    resources: { serial: "radio.serial3" },
```
gbin: if this is a linux SBC, the radio id naming is strange. It should be more like os.serial3 or linux.serial3? (so the user doesn't get confused)

components/bridges/cu_msp_bridge/README.md
```diff
@@ -30,7 +30,7 @@ bridges: [
   (
     id: "msp_bridge",
     type: "cu_msp_bridge::CuMspBridgeStd",
-    resources: { serial: "linux.serial_usb0" },
+    resources: { serial: "linux.serial3" },
```
gbin, line 25: ha here is it ok :)

components/bridges/cu_msp_bridge/README.md
```diff
@@ -6,13 +6,13 @@ Bridge that combines the MSP source/sink tasks into one serial transport. It exp
 
 The bridge expects a `serial` resource. For std targets, use
 `cu_linux_resources::LinuxResources` as the provider. It opens serial devices
-and exposes fixed serial slots such as `<bundle>.serial_usb0`.
+and exposes fixed serial slots such as `<bundle>.serial0` through `<bundle>.serial5`.
 
-For `serial_usb0`, configure serial slot `serial3_*` in the Linux bundle:
+For example, to use `serial3`, configure serial slot `serial3_*` in the Linux bundle:
```
gbin: maybe explain that bundle once in the linux resource bundle and make the readmes of the crates point to it with a link?

components/bridges/cu_msp_bridge/README.md
makeecat: what do you suggest to do here? Is that an acknowledgment that this one is good here?

components/bridges/cu_msp_bridge/README.md
makeecat: It has already been explained the `cu_linux_resources/README.md`. Now we removed the duplicated documentation about serial configuration from `cu_msp_bridge` and `cu_crsf`, and link them to the `cu_linux_resources/README.md`.

components/bridges/cu_msp_bridge/README.md
gbin: yep the id is linux, which makes sense in the context :) but still move it to the common doc form the bundle

PR #860, opened 2026-02-24
safetymon: adds configurable decision making on locks, panics & shutdown

gbin:
## Summary

We need a way in Copper to deal with fatal error like a lock, a panic or a monitor tripping with shutdown.

1. this adds a way to add several monitors to your project.
2. this "muxes" the decision from those monitors and take the safest one
3. we added a safety monitor for std that just bails the process with a configurable copperlist timeout + a list of process error codes for an external watchdog to get it from there (for example restarting a simpler. safe-mode mission in case of emergency)

## Related issues
- Closes #

## Testing
- [x] `just fmt`
- [x] `just lint`
- [x] `just test`
- [ ] optional full `just std-ci` (if std/runtime paths are impacted)
- [ ] optional full `just nostd-ci` (if embedded/no_std paths are impacted)
- [ ] Other (please specify):

pro-tip: `just` with no parameters in the root defaults to `just fmt`, `just lint`, and `just test`.

## Checklist
- [ ] I have added or updated tests where needed
- [ ] I have considered platform impact (Linux/macOS/Windows/embedded)
- [ ] I have considered config/logging changes (if applicable)
- [ ] This change is not a breaking change (or I documented it below)

### DIFF
--- a/components/monitors/cu_safetymon/tests/safety_monitoring.rs
+++ b/components/monitors/cu_safetymon/tests/safety_monitoring.rs
@@ -66,20 +66,20 @@ fn cu_logmon_and_cu_safetymon_can_run_together() {
     let probe = Arc::new(RuntimeExecutionProbe::default());
     monitors.set_execution_probe(probe.clone());
 
-    let clock = RobotClock::new();
+    let (ctx, _clock_control) = CuContext::new_mock_clock();
     let meta = CuMsgMetadata::default();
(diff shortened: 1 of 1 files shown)


makeecat (approved):
This is a breaking change. We need to note it in CHANGELOG.
Legacy monitor: configs are now silently dropped, resulting in no monitor being installed.

Apart from that, LGTM

PR #887, opened 2026-03-03
Add JointState payload and ROS adapter

NRdrgz:
## Summary
Introduce JointState payload and adds its ROS2 adapter
Change cu_feetech to make use of this payload

## Related issues
- Closes #

## Testing
- [X] `just fmt`
- [X] `just lint`
- [X] `just test`
- [ ] optional full `just std-ci` (if std/runtime paths are impacted)
- [ ] optional full `just nostd-ci` (if embedded/no_std paths are impacted)
- [ ] Other (please specify):

pro-tip: `just` with no parameters in the root defaults to `just fmt`, `just lint`, and `just test`.

## Checklist
- [X] I have updated docs or examples where needed
- [X] I have added or updated tests where needed
- [X] I have considered platform impact (Linux/macOS/Windows/embedded)
- [X] I have considered config/logging changes (if applicable)
- [ ] This change is not a breaking change (or I documented it below)

## Breaking changes (if any)
Breaking change for people using cu_feetech, relatively new yet

### DIFF
--- a/components/payloads/cu_sensor_payloads/src/joint_state.rs
+++ b/components/payloads/cu_sensor_payloads/src/joint_state.rs
@@ -31,6 +31,18 @@ impl<const N: usize> JointState<N> {
     }
 }
 
+impl<const N: usize> std::fmt::Display for JointState<N> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "pos:{:?} vel:{:?} eff:{:?}",
+            self.positions.as_slice(),
+            self.velocities.as_slice(),
+            self.efforts.as_slice()
+        )
+    }
+}
+
 // CuArray only implements Decode<()> (unit context), not the generic
 // Decode<__Context> that the derive macro would require.  Mirror the same
 // manual pattern used by CuArray itself.
--- a/examples/cu_feetech_demo/src/main.rs
+++ b/examples/cu_feetech_demo/src/main.rs
@@ -117,8 +117,8 @@ mod tasks {
         }
 
         fn process(&mut self, _ctx: &CuContext, input: &Self::Input<'_>) -> CuResult<()> {
-            if let Some(js) = input.payload() {
-                debug!("Present positions: {:?}", js.positions.as_slice());
+            if let Some(positions) = input.payload() {
+                debug!("Present positions: {}", positions);
             }
             Ok(())
         }


components/payloads/cu_sensor_payloads/src/joint_state.rs
```diff
@@ -0,0 +1,104 @@
...
+///
+/// `N` is the maximum number of joints stored in the fixed-capacity arrays.
+/// Choose it to match your robot; `8` covers most 6-DOF arms with a spare.
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Encode, Reflect)]
+#[reflect(from_reflect = false, no_field_bounds)]
+pub struct JointState<const N: usize> {
+    /// Joint names matching the ROS 2 `sensor_msgs/JointState.name` field.
+    /// Empty when the producer does not have URDF name information.
+    pub names: CuArray<String, N>,
```
gbin: We should avoid heap allocated, id in copper messages, it breaks the linearity of the copperlist.

components/payloads/cu_sensor_payloads/src/joint_state.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Joint positions (radians for revolute joints, metres for prismatic).
+    pub positions: CuArray<f32, N>,
```
gbin, line 21: please use cu-units.

components/payloads/cu_sensor_payloads/src/joint_state.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Joint velocities (rad/s or m/s).
+    pub velocities: CuArray<f32, N>,
```
gbin, line 23: same, cu-units

components/payloads/cu_sensor_payloads/src/joint_state.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Joint efforts (N·m for revolute joints, N for prismatic).
+    pub efforts: CuArray<f32, N>,
```
gbin, line 25: cu-units

components/payloads/cu_sensor_payloads/src/joint_state.rs
gbin, line 21: also, can you check the transforms? maybe we can make them compatible / composable?

components/payloads/cu_sensor_payloads/src/joint_state.rs
```diff
@@ -0,0 +1,104 @@
...
+/// outputs radians, degrees, raw ticks, or normalised values.
+///
+/// # Type parameter
+///
+/// `N` is the maximum number of joints stored in the fixed-capacity arrays.
+/// Choose it to match your robot; `8` covers most 6-DOF arms with a spare.
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Encode, Reflect)]
+#[reflect(from_reflect = false, no_field_bounds)]
+pub struct JointState<const N: usize> {
```
gbin, line 19: It looks like it has been shaped as an SOA this structure, maybe use our SOA facility so people can get an API automatically like bulk transforms and get specific joint

examples/cu_feetech_demo/src/main.rs
```diff
@@ -117,8 +117,8 @@ mod tasks {
         }
 
         fn process(&mut self, _ctx: &CuContext, input: &Self::Input<'_>) -> CuResult<()> {
-            if let Some(positions) = input.payload() {
-                debug!("Present positions: {}", positions);
+            if let Some(js) = input.payload() {
+                debug!("Present positions: {:?}", js.positions.as_slice());
```
gbin: the logs in copper doesn't support :? yet (it could be a good addition at some point, I made the same mistake so many times too lol)

components/payloads/cu_sensor_payloads/src/joint_state.rs
gbin: Also in general, we prefer "hard" IDs, for example for macros so if you try to use a message and typoed an ID it is not crashing on your robot but fails to compile instead.  Check transforms for that I think we end up with a good compile time structure over there.

components/payloads/cu_sensor_payloads/src/joint_state.rs
NRdrgz: Positions can either be rad or meters depending on the type of joint, can I use cu-unit for this?

components/payloads/cu_sensor_payloads/src/joint_state.rs
NRdrgz: What do you think about removing names from the Copper JointState payload and just have it in the ROS adapter? And define the joint names from the .ron file when using the ROS2 bridge?

NRdrgz: Or leave names as empty in the ROS adapter as well and let users implement their own version of the ROS adapter with the joint names?

components/payloads/cu_sensor_payloads/src/joint_state.rs
gbin: with cu29-units you will put just what you want: Angle, Velocity, etc... then when you pick the value, you ask it in the unit you want: be it league/day or m/s :)

```rust
// Standardized barometer payload carrying pressure and temperature.
#[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize, Encode, Decode, Reflect)]
pub struct BarometerPayload {
    pub pressure: Pressure,
    pub temperature: ThermodynamicTemperature,
}
```

then if you want a specific unit:
```rust
pressure.get::<pascal>()
```

all typed, really powerful. You divide Length by Time you get Velocity etc etc..

gbin: https://github.com/copper-project/copper-rs/tree/master/components/libs/cu_transform

this lib

components/payloads/cu_sensor_payloads/src/joint_state.rs
NRdrgz: I've tried many things and I think what makes more sense is just to remove names from the JointState copper payload and have the ROS adapter fill default jointX names. Done in last commit

gbin: Yeah I feel this could be some kind of config on the ROS adapter structure indeed. This sounds natural because Copper considers a robot *static*, will you add a joint when the robot is moving? probably never ... so why allocate / deallocate / stringly compare at 1khz, every, single, joint, that, never, ever, change lol :)

We need to try very hard to not let the very bad ROS design choices infect copper this is why we need a little bit of heavy lifting on the adaptor side.

You are building here an SOA, I would use this: https://github.com/copper-project/copper-rs/blob/master/core/cu29_soa_derive/tests/proctest.rs

so you get all the accessors etc ..

NRdrgz: "We need to try very hard to not let the very bad ROS design choices infect copper"
100% agree
Taking a look at SOA implementation + cu unit, thanks

examples/cu_feetech_demo/src/main.rs
NRdrgz: Thank you, fixed it

PR #1076, opened 2026-05-06
Fix missing external example references after PR #1053

neilmacintyre:
## Summary
Add links to moved heavy examples in README/AGENTS/component docs and replace stale local paths with the corresponding extra-examples GitHub URLs introduced by PR #1053.

## Reminder

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -129,10 +129,10 @@ Use these as the quickest way to regain context:
 - `examples/cu_caterpillar/`
   - Minimal but realistic Copper app.
   - Good reference for `#[copper_runtime]`, logger setup, determinism, resim, and justfile helpers.
-- `examples/cu_flight_controller/`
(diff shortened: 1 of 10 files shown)


neilmacintyre:
I'm running into lots of missing dependencies when running `just`. Is there an install script I should run before running `just`?

gbin:
@neilmacintyre thanks for the fixes!

I tried to have those shortcuts tell you what to install, what does it tell you now? 
Maybe we could bundle them in one install step.

neilmacintyre:
`clang`, `libclang-dev`, `taplo-cli `, `fmtron`, `cargo-deny`,`typos-cli` and `sccache` were all missing on my machine. Its worth noting that all these dependencies are installed in the image produced by [Dockerfile.ubuntu](https://github.com/copper-project/copper-rs/blob/master/support/docker/Dockerfile.ubuntu).

I'm building a devcontainer with `Dockerfile.ubuntu` as the base. If this is something that you think could be useful I can make a PR for it [edit first pass here: https://github.com/neilmacintyre/copper-rs/pull/1]

PR #1102, opened 2026-05-16
fix: capture arguments in logging

Scofield626:
## Summary

  Fix structured text log formatting when a log message mixes named placeholders like `{hash}` with positional placeholders like `{}`.

  ## Related issues
  - Closes #962

  ## Changes

  - Replace named placeholders in the existing positional-formatting path for rebuilt text logs.
  - Apply the same minimal formatting fix to debug live-text log formatting.
  - Correct debug live-text named parameter mapping so anonymous parameters are not accidentally paired with named parameter labels.
  - Add regression tests for mixed named and positional placeholders.

  ## Reminder
  - [x] I ran `just` from the repo root


gbin:
ideally I would love to have this:

debug!("Hash: {hash}");

for a named parameter instead of this:
debug!("Hash: {hash}", hash = hash); // <- I think this change basically would require that in the test no?

Scofield626:
> ideally I would love to have this:
> 
> debug!("Hash: {hash}");
> 
> for a named parameter instead of this: debug!("Hash: {hash}", hash = hash); // <- I think this change basically would require that in the test no?

Good call — done. The macros now do std-style implicit capture. Inline` {identifier}` (and `{identifier:?}`) placeholders are captured from the caller's scope and logged as named params. Explicit `name = value` still works and takes precedence if both are present (so `debug!("{x}", x = other`) does what you'd expect), and positional `{}` is unchanged.

gbin (approved):

gbin:
Thx

gbin:
The failure on CI is unrelated, I am working on it.

elpiel:
Thank you for working on this @Scofield626 !

PR #1107, opened 2026-05-19
fix: order of args in resouces macro

Scofield626:
## Summary

`resources!` produced code whose compilability depended on the order entries was declared: an `Owned<_>` entry placed after any `Shared<_>/Borrowed<_>` entry failed with E0502: cannot borrow value as mutable because it is also borrowed as immutable, while the same entries in a different order compiled fine.

**Root cause:** the macro generated `from_bindings` as a single struct literal, and struct-literal fields are evaluated in the user's declaration order. The accessors borrow asymmetrically: `borrow<'r, T>(&'r self) -> Borrowed<'r, T>` holds an immutable borrow of the manager for the whole `'r`, while `take<T>(&mut self)` needs a transient mutable borrow. If a shared borrow was evaluated before an owned take, the 'r-long immutable borrow was still live when take requested `&mut`.

**Fix**: decouple fetch evaluation order from declaration order. The generated struct Resources keeps fields in the user's declared order, but `from_bindings` now emits the fetches as ordered let bindings, i.e., all owned takes first, then all shared/borrowed borrows, then assembles the struct via field-init shorthand. Owned takes only need a transient &mut, so doing them all before any 'r shared borrow is the one ordering the borrow checker accepts, for every declaration order. Resources are independently keyed, so reordering fetches is semantically safe.

## Related issues

- Closes #956 

## Changes

- core/cu29_derive/src/resources.rs:
  - Bucket per-entry fetch statements into owned (take) and shared/borrowed (borrow) groups; emit all owned takes before any borrow in from_bindings.
  - Struct field order remains the user's declared order (stable public API and Debug).
  - Collapse the duplicated from_bindings body across the two needs_lifetime arms into one shared from_bindings_body token stream; only the impl header/manager mutability differ now.
- core/cu29_derive/tests/compile_pass/resources/owned_after_shared.rs: new trybuild regression test covering shared-first (the failing snippet from the issue), owned-first, and interleaved orderings.


gbin (approved):

gbin:
Perfect thanks

PR #1122, opened 2026-05-27
feat: only log what you use for CuHandle

### LINKED ISSUE #1082: feature: "only log what you use" Add an option to not log CuHandles that never got read
## Problem statement

Typical case: you have a 120fps camera but you can only infer at 10fps, you can ruin your logging bandwidth for 90+% of images that won't get even touched by your robot.

The current workaround: disable logging for the source and log the image alongside the inference when it happens.

## Proposed solution

Make an option on the video source that use a special handle with a "logme" / "touched" boolean and adapt the serialization of the CuHandle to just log the metadata and not the content (we will know a frame occured at that time at least).

## Impact / scope (optional)
- Affected crates, platforms, or features:

Scofield626:
## Summary

Adds an opt-in "only log what you use" mode for CuHandle-bearing payloads: high-rate sources can be configured to skip payload bytes for frames no downstream consumer ever read, while preserving timestamps/metadata in the unified log.

## Related issues

- Closes #1082

## Changes

- **`CuHandle`** wraps `Arc<CuHandleCell<T>>` with `AtomicBool` (touched) + `AtomicU8` (mode). New API: `mark_touched`, `with_touched_inner`, `payload_should_log`, `apply_handle_content_policy`.
- **`HandleContent` enum** (`all` / `touched_only` / `none`) under `NodeLogging.handle_content`. RON: `logging: (handle_content: touched_only)`.
- **Per-slot codegen** stamps the policy on payload handles before encoding and routes to a new `encode_metadata_only` helper when bytes should be skipped. Zero codegen change for default `all`.
- **Safety gate**: non-default `handle_content` requires `impl HandleContentAware for MyPayload {}`. Otherwise it's a compile error instead of a silent no-op.
- **`CuImage`** forwards the policy methods to its inner handle and implements `HandleContentAware`.
- **`cu_apriltag`** calls `payload.mark_touched()` where it reads the frame.
- **Wire format unchanged** — skipped frames reuse the existing `0u8` no-payload tag.
- **Tests**: 11 pool unit tests, 2 encoder tests, 2 config round-trips, 5 end-to-end `CuImage` tests, 1 `compile_pass` + 1 `compile_fail`.


gbin:
Sorry for the delay on this one, I was pondering about some details but I think it is good enough, we can always extend this in the future. Thanks for your contribution!

PR #1128, opened 2026-06-05
Fix cu-tuimon hostname lookup on Windows

knuttu94:
## Summary
Fixes hostname retrieval in `cu-tuimon` for Windows/native builds.

The previous `native` implementation used `libc::gethostname`, which assumes a
Unix-style libc path. That works on Unix-like systems, but it does not match the
broader set of native platforms covered by Copper's `native` build path and
caused build issues on Windows.

## Changes
This change does the following:
- Replaces the `native` hostname lookup with the [`hostname`](https://crates.io/crates/hostname) crate
- Keeps the existing browser implementation unchanged
- Includes `hostname` only for native targets

## Reason
`cu-tuimon` currently has separate `native` and `browser` code paths.
The previous `native` implementation assumed a Unix-specific hostname API for all native targets. 
Using `hostname`, a cross-platform hostname crate, preserves the structure while avoiding platform-specific handling of hostname inside `cu-tuimon`.

## Validation
I validated the implementation with:
- `cargo check -p cu-tuimon`
- `cargo check -p cu-tuimon --target wasm32-unknown-unknown`

## Reminder

### DIFF
--- a/components/libs/cu_tuimon/Cargo.toml
+++ b/components/libs/cu_tuimon/Cargo.toml
@@ -43,8 +43,8 @@ tui-widgets = { version = "0.7.0", default-features = false, features = [
 web-time = "1.1.0"
 
 [target.'cfg(not(all(target_family = "wasm", target_os = "unknown")))'.dependencies]
+hostname = "0.4.2"
 dotenvy = { version = "0.15.7", optional = true }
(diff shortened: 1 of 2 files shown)


gbin:
Thx for the contrib!

PR #1134, opened 2026-06-15
thread pools with CPU affinity & real-time scheduling

Scofield626:
### Summary

  Thread pools were previously a minimal resource-layer concept: a single auto-injected`ThreadPoolBundle` with one knob (`config: {"threads": N}`), and the `parallel-rt` executor had no real per-stage worker control. This PR makes thread pools **first-class, named, and declarative**, each carrying an optional **CPU affinity** and **real-time scheduler policy/priority**, consumed by both the background-task pools and the `parallel-rt` execution engine.

  The feature is static/declarative (pools known at compile time, wired by the macro) and **`std`-only** `no_std`/embedded targets ignore it entirely.

  ### What's new

  **Config schema (`runtime.thread_pools`)**

  ```ron
  runtime: (
      thread_pools: [
          ( id: "rt",         threads: 4, affinity: [2, 3, 4, 5], scheduler: Fifo(priority: 80) ),
          ( id: "background", threads: 2, affinity: [0, 1] ),
          ( id: "vision",     threads: 2, scheduler: Nice(10), on_error: Strict ),
      ],
  ),
  ```

  - `Scheduler`: `Other` / `Nice(i8)` / `Fifo { priority }` / `RoundRobin { priority }`
  - `OnError`: `Warn` (log and fall back to default scheduling — keeps unprivileged dev runs working) or `Strict` (hard-fail at startup, for deployed RT robots)
  - Config-time validation of ids, thread counts, priority/nice ranges, and affinity lists.

  **Pool builder** — `thread_pool::build_pool()` builds a named rayon pool and applies affinity (Spread: worker `i` → `core[i % len]`) and scheduler policy **once per worker at startup** via
  `ThreadPool::broadcast`, never on the hot path. Gated behind a new `std`-only `rt-scheduling` feature (pulls `core_affinity` + `libc`); RT policies are Linux-only, affinity is cross-platform. With the feature off, affinity/scheduler are ignored with a warning.

  **Background tasks** now run on named pools. `background: true` uses the default `"background"` pool;
  `background: (pool: "vision")` selects an explicit one. Pool names resolve to registry indices at macro-expansion time — an unknown pool is a compile error.

  **parallel-rt** — each stage worker pins itself and sets its scheduler from the reserved `"rt"` pool at startup. On a `Strict` pool, failure aborts the pipeline.

  ### Testing

  - Config: RON parse/round-trip, both `background` forms, validation failures; `std` + `no_std` build.
  - Builder: thread count, `Warn` tolerance, `Strict`-on-invalid-affinity, valid-core pinning; clippy
  clean with and without `rt-scheduling`.
  - Runtime: default-pool injection, legacy migration; derive + runtime suites green.
  - Determinism regression across missions.
  - API v1 baselines refreshed for the new config types and `thread_pool` module.


gbin:
This is pretty good!

Originally I thought putting those pools in resource bundles but it looks better in the runtime config for sure. Have you removed my old code from the resources? 

Another thing, can you open a PR on the copper book to see how the documentation would look like?

And finally, example wise/benchmark wise, any specific example we can demo that with?

Scofield626:
Hey @gbin,

The old code is not removed yet. The original `ThreadPoolBundle` code is still in core/cu29_runtime/src/resource.rs, fully intact. That includes the ThreadPoolBundle struct, ThreadPoolId::BgThreads, and the build() that spins up the rayon pool via ThreadPoolBuilder (resource.rs:437–491). I deliberately kept it rather than deleted it, but I should be precise about its status: it's now effectively dead code. Shall I keep this for the moment or remove in this PR?


> Another thing, can you open a PR on the copper book to see how the documentation would look like?
And finally, example wise/benchmark wise, any specific example we can demo that with?

I'll look into those you mentioned.

Scofield626:
After running benchmarks on https://github.com/copper-project/copper-rs/tree/master/examples/cu_parallel_mandelbrot with affinity pinning, it shows consistent performance as the version without defining explicit threadpool. 

Also, one thing to notice is that this PR introduced some unsafe code for enabled some libc scheduler policy, please take a look and let me know what you think

core/cu29_derive/src/lib.rs
```diff
@@ -1640,7 +1640,7 @@ pub fn copper_runtime(args: TokenStream, input: TokenStream) -> TokenStream {
             let threadpool_bundle_index = if task_specs.background_flags.iter().any(|&flag| flag) {
                 match bundle_specs
                     .iter()
-                    .position(|bundle| bundle.id == "threadpool")
+                    .position(|bundle| bundle.id == THREAD_POOL_BUNDLE_ID)
```
gbin: as commneted offline, if we drop the pool from the resources, let's do that completely

core/cu29_derive/src/lib.rs
```diff
@@ -2102,6 +2102,51 @@ pub fn copper_runtime(args: TokenStream, input: TokenStream) -> TokenStream {
...
+                    .enumerate()
+                    .map(|(index, pool)| (pool.id.as_str(), index))
+                    .collect()
+            })
+            .unwrap_or_default();
+        for (task_index, pool_name) in task_specs.background_pools.iter().enumerate() {
+            if !task_specs.background_flags[task_index] {
+                continue;
+            }
+            // "rt" is reserved for the parallel-rt engine and is never task-bound.
```
gbin: the better explanation here is that rt is the default pool for all the tasks EXCEPT those where we override it from the config with background

core/cu29_derive/src/lib.rs
```diff
@@ -7102,14 +7188,51 @@ fn build_resources_module(
...
+                        for (pool_index, pool_spec) in runtime.thread_pools.iter().enumerate() {
+                            // The reserved "rt" pool is applied to the parallel-rt
+                            // stage workers directly, not borrowed as a rayon pool;
+                            // skip it so we don't spawn idle workers. Its slot stays
+                            // unfilled and is never looked up.
+                            if pool_spec.id == cu29::config::RT_POOL {
+                                continue;
+                            }
+                            let pool = cu29::thread_pool::build_pool(pool_spec)?;
+                            let pool_key = cu29::resource::ResourceKey::<()>::new(
```
gbin: same: let's completely decouple from the resources. We will revisit that maybe someday for NUMA and CPU types.

gbin (changes requested):
main thing: remove all the resources interactions otherwise this is pretty good!

gbin (approved):

PR #1135, opened 2026-06-19
feat: monitor for custom memory (heap) allocator

Scofield626:
## Problem

`core/cu29_runtime/src/monitoring.rs` already ships a `CountingAlloc` global allocator and a `ScopedAllocCounter` scope helper behind the `memory_monitoring` feature, but the feature is dead:

- No call site in the repo ever constructs a `ScopedAllocCounter`.
- Its `Drop` impl computes the delta and drops it on the floor (`TODO(gbin)`,`monitoring.rs:1945-1950`).
- The `CuMonitor` trait only sees post-copperlist views (`CuMsgMetadata` timings) — there is no hook for memory deltas, and no per-task lifecycle observation.

The maintainer wants users to know, in realtime-sensitive code, whether any of their tasks allocate on the heap between `start` and `stop`, and ideally to get heap-balance stats per lifecycle.

## Why "monitor + debug-only feature" is awkward

Two link-time constraints:

1. `#[global_allocator]` is a whole-binary decision. The counting allocator must live in `cu29_runtime` (or be opted in by the user's `main.rs`); a monitor selected by RON at runtime cannot install one.
2. Per-task allocation accounting must come from the proc-macro (`cu29_derive/src/lib.rs`) wrapping each generated `task.start/preprocess/process/postprocess/stop` call — the monitor cannot reach into those sites itself.

## Decision

- Keep allocator installation gated behind `memory_monitoring` → **zero cost when off**.
- Gate the macro emission of allocation scopes behind the same `cfg!(feature = "memory_monitoring")` → no overhead on default builds.
- Make `cu_memmon` a normal monitor crate sibling to `cu_logmon`/`cu_safetymon`, added via RON. When the feature is off it warns once at `start` and reports zeros; users never need `#[cfg]` in their code.
- `realtime_strict` mode is **opt-in via RON** — too dangerous as a default for users who allocate on purpose.
- Defer unifiedlog `MemSample` encoding until someone asks.

## Changes
1. **Expose counters** in `cu29_runtime::monitoring`: `global_allocated_bytes() / global_deallocated_bytes() -> Option<usize>` (`None` when the feature is off). Keep `CountingAlloc` `pub` for users who want to wrap their own allocator.
2. **Add `CuMonitor::observe_alloc`** with a default no-op: `fn observe_alloc(&self, _component_id: ComponentId, _step: CuComponentState, _allocated: usize, _deallocated: usize) {}`. Tuple impl fans out. Existing monitors compile unchanged.
3. **Wire the macro** (`cu29_derive/src/lib.rs`):
   - Wrap each `task.<step>()` call in a `ScopedAllocCounter`.
   - On return, forward the delta via `monitor.observe_alloc(...)`.
   - Gated on `cfg!(feature = "memory_monitoring")`.
4. **`components/monitors/cu_memmon`** (mirrors `cu_logmon` shape):
   - `observe_alloc`: per-`(component, step)` running totals + per-component lifetime balance since `start` returned.
   - `process_copperlist`: throttled `info!` one-liner, e.g. `[mem] foo start=+4KiB -0B | proc max=+0B | stop=+0B -4KiB | leak=0B`.
   - `stop`: final per-task `leak = alloc - dealloc` report. `warning!` on non-zero.
   - RON knob `realtime_strict: bool` → any non-zero alloc in `preprocess/process/postprocess` returns `Decision::Shutdown`.
   - Feature-off path: emit one `warning!` at `start` and stop there.

### DIFF
--- a/.github/actions/rust-ci-setup/action.yml
+++ b/.github/actions/rust-ci-setup/action.yml
@@ -107,6 +107,17 @@ runs:
       if: inputs.install_nextest == 'true'
       uses: taiki-e/install-action@nextest
 
+    # The justfile uses just's `shell(...)` function which was added in 1.27.0.
+    # Linux host runners get just 1.21 from apt, which fails to parse the
+    # justfile. Install a current just binary from the upstream release feed;
+    # taiki places it in ~/.cargo/bin which precedes /usr/bin on PATH so this
+    # also overrides any apt-installed just inside prebaked images.
+    - name: Install recent just (Linux)
+      if: runner.os == 'Linux'
+      uses: taiki-e/install-action@v2
+      with:
+        tool: just
+
     - name: Set workspace scope
       if: inputs.set_workspace_flag == 'true'
       shell: bash
--- a/.github/actions/rust-ci-setup/action.yml
+++ b/.github/actions/rust-ci-setup/action.yml
@@ -107,17 +107,6 @@ runs:
       if: inputs.install_nextest == 'true'
       uses: taiki-e/install-action@nextest
 
-    # The justfile uses just's `shell(...)` function which was added in 1.27.0.
-    # Linux host runners get just 1.21 from apt, which fails to parse the
-    # justfile. Install a current just binary from the upstream release feed;
-    # taiki places it in ~/.cargo/bin which precedes /usr/bin on PATH so this
-    # also overrides any apt-installed just inside prebaked images.
-    - name: Install recent just (Linux)
-      if: runner.os == 'Linux'
-      uses: taiki-e/install-action@v2
-      with:
-        tool: just
-
     - name: Set workspace scope
       if: inputs.set_workspace_flag == 'true'
       shell: bash
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ members = [
   "components/monitors/cu_consolemon",
   "components/monitors/cu_bevymon",
   "components/monitors/cu_logmon",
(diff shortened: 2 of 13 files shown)


gbin:
I would like to pull the trigger for v1. Would this introduce backward compatibility issue if we stabilize the API?

Scofield626:
I don't think anything violates the backward compatibility. No public items are removed, renamed, or have their signatures changed. Are there anything particular raising your concerns on that? 

I think this is also directly validated by `just api-check`?

gbin:
yep just api-check is the test to see what was added. Is it still a draft?

.github/actions/rust-ci-setup/action.yml
```diff
@@ -145,6 +145,17 @@ runs:
       if: inputs.install_nextest == 'true'
       uses: taiki-e/install-action@nextest
 
+    # The justfile uses just's `shell(...)` function which was added in 1.27.0.
```
gbin, line 110: I believe I had to do that in a previous PR (ie install just for ci/cd)

gbin (approved):
Maybe install the monitor on one of the main example like balancebot? So we exercise it regularly

.github/actions/rust-ci-setup/action.yml
Scofield626: after rebasing, removing these have made cicd failed again for the same reason @gbin

gbin:
@Scofield626 ok ok can you put the ci change quickly in another pr, we land it then land this one? thanks

PR #1170, opened 2026-06-30
Chore: latest nightly has a slight API update diff

gbin:
## Related issues
- Closes #

## Reminder


gbin:
@Scofield626 you might be hit by that: you need to do "rustup update" and it should catch up

PR #1180, opened 2026-07-06
feat: anytime task support

Scofield626:
Adds the v1 anytime-task interface to `cu29_runtime`: an `Anytime` trait plus an `AnytimeTask<A>` adapter for computations that produce a usable base result and refine it within a per-cycle window.

## Scope

This PR is intentionally limited to the anytime interface and adapter. It does not change `CuContext`, the existing task traits, simulation overrides, or generated runtime execution.

- `Anytime` exposes `base`, bounded `refine`, and in-place `write_best`, while mirroring normal task lifecycle and `Freezable` behavior.
- `AnytimeTask<A>` stores only the inner task and refinement duration. It owns no output cache and performs no adapter-owned output clone.
- `AnytimeOutput<O>` carries the best value plus `Progress`, distinguishing refinement-budget exhaustion, convergence, and task-defined satisfaction.
- The required RON field is `refinement_budget_us` and must be positive. There is no compatibility alias for `budget_us`.
- Missing input always clears the output payload and sets `tov` to `Tov::None`.

## Refinement window

`refinement_budget_us` applies only to starting refinements:

- `base` runs before the window starts.
- The adapter measures the window with physical `CuInstant`, independently of `RobotClock`.
- No refinement starts at or after the deadline.
- A refinement already in progress may finish after the deadline.
- `write_best` runs after the window.

This is cooperative timing: base work, one already-started refinement, and output commitment are not hard-bounded by this setting.

## Replay

Exact-output replay continues to use Copper’s existing path, which restores the recorded output and task keyframe without executing the task; the integration test covers that behavior. Deterministic debug re-execution by recorded refinement count is intentionally outside this PR and would require separate runtime/codegen design.

## Verification

- `just fmt`
- focused anytime unit and exact-output replay tests
- `just api-check`
- `just pr-check`
- `just nostd-ci`


core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,964 @@
...
+    /// Return a `CuError` from `process`; the existing copper-rs error path
+    /// takes over.
+    Fail,
+}
+
+impl OnOverload {
+    fn parse(s: &str) -> CuResult<Self> {
+        match s {
+            "reuse_last" | "ReuseLast" => Ok(Self::ReuseLast),
+            "fail" | "Fail" => Ok(Self::Fail),
```
gbin: Why do we need this stringly type thing?

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,964 @@
...
+    ReuseLast,
+    /// Return a `CuError` from `process`; the existing copper-rs error path
+    /// takes over.
+    Fail,
+}
+
+impl OnOverload {
+    fn parse(s: &str) -> CuResult<Self> {
+        match s {
+            "reuse_last" | "ReuseLast" => Ok(Self::ReuseLast),
```
gbin: We are defining right now the API, why it needs to be fuzzy?!

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,964 @@
...
+    /// Cheap, MUST always fit within any plausible budget, produces a safe
+    /// rough result. Reads `input` and resets per-cycle scratch on `self`.
+    /// After this returns, `best()` and `progress()` are valid.
+    fn base(&mut self, input: &Self::Input, ctx: &CuContext) -> CuResult<()>;
+
+    /// One bounded refinement increment, sized so overshoot is at most one
+    /// step. MUST NOT allocate on the RT path. Signature deliberately takes no
+    /// `ctx` — `refine` must be pure w.r.t. wall time; the adapter checks the
+    /// deadline between calls.
+    fn refine(&mut self) -> CuResult<Step>;
```
gbin: why not resending input? This forces the task to maybe store it why it is pending in the copperlist.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,964 @@
...
+
+    /// Best-so-far. Cheap. Valid after `base` and after every `refine`.
+    fn best(&self) -> Self::Output;
```
gbin: the taste of this API is not really like the copper-rs flavor, it forces a copy of Output -> this is not good. Look at every single other task API and you will see we *give* a mutable output to avoid any copies.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,964 @@
+//! Anytime tasks: interruptible computations that produce a usable base result
```
gbin, line 1: ok I stopped reviewing this is better than the original idea but not there yet, let harp on it

gbin (changes requested):
Please don't rely too much on LLMs for API definition for copper, they are notoriously bad at it.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+    }
+    fn postprocess(&mut self, _ctx: &CuContext) -> CuResult<()> {
+        Ok(())
+    }
+    fn stop(&mut self, _ctx: &CuContext) -> CuResult<()> {
+        Ok(())
+    }
+}
+
+/// Adapter that turns any `A: Anytime` into a `CuTask` visible to
```
gbin, line 140: why not first class support? An actual CuAnytimeTask? It feels like it adds layer of complexity with no real benefit here.

The runtime HAS to know it is an anytime task, like the user so simpler is better.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+    ///
+    /// Cheap, MUST always fit within any plausible budget, produces a safe
+    /// rough result. Reads `input` and resets per-cycle scratch on `self`.
+    /// Return [`Step::Continue`] when refinement can proceed, or an early-exit
+    /// status when the base result already converged or met its quality target.
+    /// After this returns, [`Self::write_best`] is valid.
+    fn base(&mut self, input: &Self::Input, ctx: &CuContext) -> CuResult<Step>;
+
+    /// One bounded refinement increment, sized so overshoot is at most one
+    /// step. MUST NOT allocate on the RT path. Signature deliberately takes no
```
gbin, line 108: no need to precise: MUST NOT allocate on the RT path, this is global to copper.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+    ///
+    /// The mutable receiver and fallible result also cover bounded accelerator
+    /// readback without adding an anytime-specific lifecycle hook. This method
+    /// runs after the refinement loop ends and must complete promptly.
+    fn write_best(&mut self, out: &mut Self::Output) -> CuResult<()>;
+
+    fn start(&mut self, _ctx: &CuContext) -> CuResult<()> {
+        Ok(())
+    }
+    fn preprocess(&mut self, _ctx: &CuContext) -> CuResult<()> {
```
gbin, line 129: the semantic of pre and post need to be precised here.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+        Self: Sized;
+
+    /// Base computation: the basic result for the anytime algorithm.
+    ///
+    /// Cheap, MUST always fit within any plausible budget, produces a safe
+    /// rough result. Reads `input` and resets per-cycle scratch on `self`.
+    /// Return [`Step::Continue`] when refinement can proceed, or an early-exit
+    /// status when the base result already converged or met its quality target.
+    /// After this returns, [`Self::write_best`] is valid.
+    fn base(&mut self, input: &Self::Input, ctx: &CuContext) -> CuResult<Step>;
```
gbin, line 105: why ctx is last?

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+
+    /// One bounded refinement increment, sized so overshoot is at most one
+    /// step. MUST NOT allocate on the RT path. Signature deliberately takes no
+    /// `ctx` — `refine` must be pure w.r.t. wall time; the adapter checks the
+    /// deadline between calls.
+    ///
+    /// `input` is re-passed each call so implementors never need to cache it
+    /// on `self` between the adapter's `base`/`refine` calls — the copperlist
+    /// already owns it for the whole cycle.
+    fn refine(&mut self, input: &Self::Input) -> CuResult<Step>;
```
gbin, line 115: how this will work if it has no ctx?

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+
+    /// Write the best-so-far into `out` in place. `out` is the runtime-owned
+    /// slot, preserved across cycles, so implementations should reuse its
+    /// allocations (for example, clear a `Vec` and extend it in place).
+    ///
+    /// The mutable receiver and fallible result also cover bounded accelerator
+    /// readback without adding an anytime-specific lifecycle hook. This method
+    /// runs after the refinement loop ends and must complete promptly.
+    fn write_best(&mut self, out: &mut Self::Output) -> CuResult<()>;
```
gbin, line 124: I think we also need ctx here

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+    /// step. MUST NOT allocate on the RT path. Signature deliberately takes no
+    /// `ctx` — `refine` must be pure w.r.t. wall time; the adapter checks the
+    /// deadline between calls.
+    ///
+    /// `input` is re-passed each call so implementors never need to cache it
+    /// on `self` between the adapter's `base`/`refine` calls — the copperlist
+    /// already owns it for the whole cycle.
+    fn refine(&mut self, input: &Self::Input) -> CuResult<Step>;
+
+    /// Write the best-so-far into `out` in place. `out` is the runtime-owned
```
gbin, line 117: I don't know if it is better to always have fn refine(&mut self, input: &Self::Input, &mut output &Self::Output) instead of a 2 calls thing.

Can we check on some algorithms. We can guarantee that we represent the same output if they need it.

Every callback, in order ops etc might require synchronization etc.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+/// ```
+///
+/// `refinement_budget_us` bounds only the window in which new refinement
+/// steps may start. The base computation runs before that window, and
+/// [`Anytime::write_best`] runs after it. A refinement that starts before the
+/// deadline is allowed to finish after it because synchronous task code cannot
+/// be safely preempted.
+#[derive(Reflect)]
+#[reflect(no_field_bounds, from_reflect = false, type_path = false)]
+pub struct AnytimeTask<A: Anytime + TypePath> {
```
gbin, line 160: this should just be owned my the runtime itself. If it is outside with no knowledge, the runtime won't be able to take any scheduling decisions on it.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,843 @@
...
+        base_step: Step,
+        mut now_nanos: impl FnMut() -> u64,
+    ) -> CuResult<Progress> {
+        match base_step {
+            Step::Converged => Ok(Progress::Converged { refinements: 0 }),
+            Step::Satisfied => Ok(Progress::Satisfied { refinements: 0 }),
+            Step::Continue => {
+                let deadline = now_nanos().saturating_add(self.refinement_budget.as_nanos());
+                let mut refinements = 0u32;
+                loop {
```
gbin, line 200: ok this is wrong. refinements should be able to be scheduled in between 2 random tasks on the copperlist with no threads. Again should be integral part of the actual scheduling, not something grafted to a cutask because a cutaskis only scheduled *once* per copperlist. 

Overall, like we discussed in person the lifecycle of this is not defined. We have to define it *clearly* when is it started, how many times per CL, monothreaded how it works, multithreaded how it works, where the dependencies plays into to copperlist, can we allow fixed scheduling?, how many copies?, do we represent the same input?, if it spread across several copper lists do we copy the input? do we copy the output?, is it premptible? do we need sync? it which situations?

gbin (changes requested):
sorry back to the drawing board, we are totally flying blind on this PR. see comments inside.

Please. LLM in read only for now. Only the design + trait, no impl.

Design doc, complete it properly:
1. what do we try to solve
2. possible solutions with all the costs, lifecycles, latencies, memory locality, std vs no_std etc etc...
3. solution selected, why?
4. tradeoffs
5. real life example of the use
6. how it doesn't paint us in a corner, future possible improvements.

PR #1181, opened 2026-07-07
Add CuRng Resource

Scofield626:
## Summary

This PR introduces a new random number generator resource with a configurable seeds, so as to support future sample-based algorithms or related things requiring rng.

It currently wraps a ChaCha8 rng for its performance and cross-platform support.

### DIFF
--- a/components/res/cu_rng/src/lib.rs
+++ b/components/res/cu_rng/src/lib.rs
@@ -50,9 +50,10 @@ pub const RNG_NAME: &str = "rng";
 
 /// Deterministic, seedable RNG resource.
 ///
-/// Backed by ChaCha8: the same seed produces the same stream on every
-/// platform and every run. This is the property replay tests and debug
-/// replay of randomized tasks rely on.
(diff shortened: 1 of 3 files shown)


components/res/cu_rng/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.inner = ChaCha8Rng::seed_from_u64(seed);
+    }
+}
+
+impl RngCore for CuRng {
+    fn next_u32(&mut self) -> u32 {
+        self.inner.next_u32()
+    }
+
+    fn next_u64(&mut self) -> u64 {
```
gbin: a lot of those algorithms won't be int based but f64 or f32, is it possible to add the support too? Maybe a more natural next::<T>() -> T or a set of type alias might be cleaner if we add 2 more types

gbin (changes requested):

gbin (approved):
just added the rand support in a prelude so it is a little bit more batteries included for the user.

PR #1182, opened 2026-07-11
fix: address Rust 1.97 Clippy lints

Scofield626:
## Summary

- remove redundant formatting borrows reported by Rust 1.97 Clippy
- iterate directly over map values in `cu-tuimon`
- use `?` to propagate structured-log reader errors

## Verification

- `just fmt`
- `just lint`
- `just api-check`
- `cargo +stable nextest run -p cu29-runtime -p cu29-export -p cu-tuimon -p cu-pid --features cu-tuimon/log_pane` (260 passed, 1 skipped)

### DIFF
--- a/components/libs/cu_tuimon/src/tui_nodes/graph.rs
+++ b/components/libs/cu_tuimon/src/tui_nodes/graph.rs
@@ -161,7 +161,7 @@ impl<'a> NodeGraph<'a> {
         // Repeat the for loop until in runs all the way through without any
         // intersections. Surely there's a more efficient way to do this.
         'outer: loop {
-            for (_, ea_them) in self.placements.iter() {
+            for ea_them in self.placements.values() {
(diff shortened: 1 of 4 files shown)


gbin (approved):

Scofield626:
@gbin  I am not sure why this is the case on API snapshot check. Could you plz take a look?

```
-pub fn cu29_runtime::config::CuConfig::render(&self, output: &mut dyn std::io::Write, mission_id: core::option::Option<&str>) -> cu29_traits::CuResult<()>
+pub fn cu29_runtime::config::CuConfig::render(&self, output: &mut dyn core::io::write::Write, mission_id: core::option::Option<&str>) -> cu29_traits::CuResult<()>
```

gbin:
This is ok. I fixed a wave of this before with the new version of the tooling. It looks like the API check gets the original name of the type instead of the alias (they are the same type)

PR #1187, opened 2026-07-12
Support multi-output source placeholders in simulation [0]

gbin:
## Summary

Allow simulation-mode applications to replace a multi-output source with a placeholder while preserving its complete output tuple. Previously the macro selected only one output type, so sources with `run_in_sim: false` could not generate the correct simulated runtime shape.

## Related issues

- None

## Changes

- Add `CuSimSrcTaskPack` as the tuple-aware placeholder for multi-output sources.
- Generate the single-output placeholder for one output and the tuple-aware placeholder for multiple outputs.
- Exercise the excluded multi-output source path in the existing compile-pass fixture.
- Update the V1 public API snapshot.

## Reminder

- [x] I ran `just` from the repo root (`just pr-check`).
- [x] I updated the module docs and API snapshot; no wiki/book changes are needed for this targeted fix.

## Additional context

Validation:

- `cargo test -p cu29-derive test_compile_fail -- --nocapture`
- `just api-check`
- `just pr-check` (638 standard tests and 329 no-default-feature tests passed)


gbin:
@Scofield626 for a quick review. I blundered this one originally hopefully we don't break existing code.

PR #1202, opened 2026-07-17
build: centralize Copper build-script setup

gbin:
## Summary
- add `cu29_build::setup()` as the single Copper build-script entry point
- replace repeated `LOG_INDEX_DIR` setup across all Copper build scripts
- update both `cargo cunew` templates

This PR intentionally preserves only the existing behavior; feature forwarding follows in the next stacked PR.

## Verification
- `just lint`
- `cargo test -p cu29-build`
- generated and built both `project` and `workspace` templates

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 
 members = [
   "core/cu29",
+  "core/cu29_build",
   "core/cu29_base_derive",
   "core/cu29_reflect_derive",
   "core/cu29_clock",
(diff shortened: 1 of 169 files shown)


gbin:
This prepares for the next one as we want to mix feature flags and compilation of the ron config grpah. It will needs another hook in the build.rs. I think this will help the user not having to worry about us hacking into the build system.

It is still backward comaptible if people do it the old fashion way this is just a helper.

Scofield626:
LGTM overall. But one minor check - shall we enable the `cu29-build` in the public-api checklist? given there seem to have some external callers?

gbin:
> LGTM overall. But one minor check - shall we enable the `cu29-build` in the public-api checklist? given there seem to have some external callers?

Good point .. is the build.rs API is part or not of the copper API? I think it is. Let me add that.

PR #1205, opened 2026-07-17
CuAnytimeTask trait for supporting anytime algorithm

Scofield626:
Adds the task-facing contract for anytime tasks: a mandatory `base()` producing a minimum valid result, then bounded `refine()` quanta the runtime schedules until its policy (time budget / quality target) says stop.

- `CuAnytimeTask`: `new`/`start`/`preprocess`/`base`/`refine`/`postprocess`/`stop`. `refine()` takes no input: `base()` captures what refinement needs into per-job state.
- `AnytimeStatus` (`Improved`/`Converged`/`Aborted`): task reports, runtime decides. Output must always hold the best result so far, so the runtime can publish at any stop point without buffering or rollback.
- `AnytimeStep`: per-quantum scheduling hints (iteration, elapsed, remaining budget, best quality).

Trait-only PR: the foreground/background runner, policy wiring, and codegen integration come separately. Includes a unit test implementing the trait end-to-end and updated `api/v1` snapshots.

### DIFF
--- a/core/cu29_runtime/src/cutask_anytime.rs
+++ b/core/cu29_runtime/src/cutask_anytime.rs
@@ -12,11 +12,13 @@ use crate::context::CuContext;
 use crate::cutask::{CuMsgPack, CuMsgPayload, Freezable};
 use crate::reflect::Reflect;
 use cu29_traits::CuResult;
+use cu29_units::si::f32::Ratio;
 
-/// Normalized quality of a published result: a unit ratio in `0.0..=1.0`, higher is
-/// better and `1.0` means no further improvement is meaningful. Sharing one scale
-/// across tasks keeps a configured quality target portable.
-pub type Quality = f32;
+/// Normalized quality of a published result: a dimensionless [`Ratio`] in
+/// `0.0..=1.0`, higher is better and `1.0` means no further improvement is
+/// meaningful. Sharing one scale across tasks keeps a configured quality target
+/// portable.
+pub type Quality = Ratio;
 
 /// Returned by [`CuAnytimeTask::base`] and [`CuAnytimeTask::refine`]; drives the
 /// runtime's refinement scheduling.
@@ -40,11 +42,6 @@ pub enum AnytimeStatus<Q> {
     /// for this job. Refinement stops; the output is published as-is, so a task that
     /// can no longer vouch even for its base result must clear the payload before
     /// returning this. The next copperlist starts a fresh job.
-    ///
-    /// This is a controlled per-job outcome, not a failure, so it carries no error:
-    /// return it when retrying with fresh input could plausibly succeed (logging any
-    /// diagnostics from the task), and return `Err` from `base`/`refine` when the
-    /// task itself is broken.
     Aborted,
 }
 
@@ -142,6 +139,7 @@ mod tests {
     use crate::cutask::CuMsg;
     use crate::input_msg;
     use crate::output_msg;
+    use cu29_units::si::ratio::ratio;
 
     /// Sums its input one increment per quantum: base publishes 0, each refine
     /// commits one more increment until the captured input is fully consumed.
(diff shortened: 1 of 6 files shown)


core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// knows the cheapest representation to retain.
+    ///
+    /// On `Err`: the job produced no valid output and the error propagates like a
+    /// `CuTask::process` failure.
+    fn base<'i, 'o>(
+        &mut self,
+        ctx: &CuContext,
+        input: &Self::Input<'i>,
+        output: &mut Self::Output<'o>,
+    ) -> CuResult<AnytimeStatus>;
```
gbin: I think either the thing errors out and pushed a CuError OR it is a controlled abort (no CuError) but having a Enum of (CuError or AnytimeStatus of  None or CuError) is just not good.
Also note: every time we have an Option or Enum this creates branching, an if statement basically, that at every single tick the CPU will have to add to the branch predictor etc... those are costly ifs and here we have 4 in a row for 99% of the time no operation at all.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,231 @@
...
+    ///
+    /// Return this when retrying with fresh input could plausibly succeed; return
+    /// `Err` from `base`/`refine` when the task itself is broken.
+    Aborted { error: Option<CuError> },
+}
+
+/// Runtime knowledge passed to [`CuAnytimeTask::refine`] — scheduling hints, not
+/// commands.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct AnytimeStep {
```
gbin: If this is for public consumption: ie our robotics task devs needs it, it should be pub, but otherwise it would be better buried in the runtime itself as a pub(crate) or probably a private struct?

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,231 @@
...
+use crate::config::ComponentConfig;
+use crate::context::CuContext;
+use crate::cutask::{CuMsgPack, CuMsgPayload, Freezable};
+use crate::reflect::Reflect;
+use cu29_clock::CuDuration;
+use cu29_traits::{CuError, CuResult};
+
+/// User-defined quality metric; higher is better. The scale is task-defined but must
+/// be consistent with the quality target configured for the task.
+pub type Quality = f32;
```
gbin: maybe a unit ratio is better here as a normalized 0 to 1?

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,231 @@
...
+///
+/// After any `Ok` return, the output must be valid and hold the best result produced
+/// so far for the current job: a quantum that regresses or plateaus keeps its
+/// candidate in task-local state and leaves the output untouched. The runtime never
+/// buffers or rolls back the output, so it can publish it at any stop point, and the
+/// published quality is monotone even when the algorithm internally is not.
+#[derive(Debug, Clone)]
+pub enum AnytimeStatus {
+    /// The output holds the best result so far; further refinement may help.
+    Improved { quality: Option<Quality> },
```
gbin: every single Option here is again a branching we will have to go through. Template the Status with a quality type or something maybe?
The rule should be: what CAN be done at compile that HAS 100% be done at compile time.

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// 0 for the first `refine()` after `base()`.
+    pub iteration: u32,
+    /// Time since `base()` started for this job.
+    pub elapsed: CuDuration,
+    /// Time left in the budget; `None` if no time budget is configured. A task can
+    /// use this to size its next quantum (e.g. skip a strategy that cannot fit).
+    pub remaining: Option<CuDuration>,
+    /// Best quality reported so far for this job, including by `base()`.
+    pub best_quality: Option<Quality>,
+}
```
gbin, line 46: I would not worry about forcing an internal struct for this right now, this can be 100% generated by the proc macro super tight, in the main loop etc. This is conceptually good but forces dynamism in the runtime.

PR #1216, opened 2026-07-21
chore: Anytime user-config

Scofield626:
## Summary

Adds the user-facing configuration surface for anytime tasks (follow-up to #1205, which introduced the `CuAnytimeTask` trait): `AnytimeConfig`, a typed `anytime:` field on `Node` next to `background:`, with six optional policy scalars organized on two axes:

| the user's question | parameter | kind |
|---|---|---|
| "How long may one result cost me?" | `time_budget_ms` | hard bound |
| "How many refinement steps may one result cost me?" | `max_refines` | hard bound |
| "When is the input **too old** for its result to be useful?" | `max_age_ms` | utility gate (age) |
| "When is the result **good enough** to stop early?" | `quality_target` | early yield |
| "When is continuing **pointless** (no progress)?" | `max_stall` | early yield |
| "When is the result **too poor** to be worth publishing?" | `quality_floor` | utility gate (quality) |

## Mapping to real-world usecases

I checked the following examples mentioned in Anytime ROS2 paper and their [repo](https://github.com/tu-dortmund-ls12-rt/Anytime-Development), which demo a good coverage of potential usage.

- RRT*: time_budget_ms, quality_floor (if no path is produced), max_stall (no progress)
- AnytimeYOLO: max_age_ms (stop when the input is too old), quality_target (score threshold), max_refines
- Monte Carlo π estimation: time_budget_ms, max_refines (iteration in their algorithm)

## Nit

I delibrately chose the `max_age_ms` instead of `deadline` for the reasons:
1. it provides a clear starting point for measurement
2. data age is a widely-used metric, so this provides better intuition, i guess?

## Test plan

- New config-parsing and validation unit tests in `config.rs` (accessor exposure, typical two-field config, each rejection path).
- `just pr-check` (fmt, lint, api-check, test) passes locally.

### DIFF
--- a/api/v1/cu29-runtime.txt
+++ b/api/v1/cu29-runtime.txt
@@ -98,6 +98,13 @@ impl cu29_runtime::config::TaskKind
 pub fn cu29_runtime::config::TaskKind::as_str(&self) -> &'static str
 impl core::convert::From<cu29_runtime::config::TaskKind> for cu29_runtime::curuntime::CuTaskType
 pub fn cu29_runtime::curuntime::CuTaskType::from(value: cu29_runtime::config::TaskKind) -> Self
+pub struct cu29_runtime::config::AnytimeConfig
+pub cu29_runtime::config::AnytimeConfig::max_age_ms: core::option::Option<f64>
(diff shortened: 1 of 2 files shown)


gbin:
perfect

PR #1226, opened 2026-07-25
Initialize CopperLists safely and clear generated slots on reuse

gbin:
## Why this is a draft

The first commit is an independent safety fix. The second commit deliberately isolates the policy question of whether every generated CopperList slot should be reset to `Default` on reuse. Please review those separately; the reset commit can be dropped without losing the initialization fix.

## Commits

1. initialize CopperLists field-by-field in uninitialized heap storage, avoiding invalid zeroed values and large stack temporaries while preserving existing reuse behavior
2. reset each generated message slot to `Default` on reuse, dropping stale payloads

## Evidence

- generated-runtime test proves a reused slot has no stale payload
- drop-tracking test proves reset releases the prior payload
- manual release microbenchmark is committed as an ignored test
- local synthetic result: preserve 0.28 ns/op, reset 0.37 ns/op, delta 0.09 ns/op

The benchmark uses a minimal one-slot payload and is only a lower-bound microbenchmark. Real graphs with large inline payloads need representative measurements before merging the second commit.

## Verification

- `just pr-check`
- 692 std tests and 375 no_std test-profile tests passed
- embedded/no_std clippy passed
- `just api-check`


gbin:
Will revisit that later with proper benchmarks, this is not strictly necessary.

PR #1227, opened 2026-07-25
Keep nested logger runtime destinations alive

gbin:
## Summary

- track logger destinations as an ID-addressed stack
- route logs to the newest live runtime destination
- flush and remove only the destination owned by the runtime being dropped
- preserve an older destination when a newer runtime is dropped
- cover both out-of-order and stack-order drops

This is the logger-lifetime fix split from the mixed debug branch and is independent of the remote-debug changes.

## Verification

- `cargo test -p cu29-log-runtime`
- std and no_std clippy with warnings denied
- `just api-check`


gbin:
Superseded by #1229. The remote debug server now enforces one active replay-session lease with heartbeat expiry, so overlapping logger destinations are no longer needed.

PR #1231, opened 2026-07-26
Add CuDepthMap sensor payload with typed length samples

gbin:
## Summary

- add CuDepthMapFormat with explicit width, height, and element stride
- add the handle-backed CuDepthMap payload using cu29-units Length elements instead of untyped f32 values
- provide typed indexed access plus an explicit meter conversion helper
- keep execution identity in CopperList metadata instead of duplicating it with a payload sequence field
- preserve padded rows, non-finite measurements, serialization, and handle-content logging policy

## Verification

- just pr-check
- cargo test -p cu-sensor-payloads --features reflect
- cargo clippy -p cu-sensor-payloads --all-targets --features reflect -- -D warnings
- cargo check -p cu-sensor-payloads --no-default-features

### DIFF
--- a/components/payloads/cu_sensor_payloads/src/depth.rs
+++ b/components/payloads/cu_sensor_payloads/src/depth.rs
@@ -1,7 +1,9 @@
 use bincode::de::Decoder;
-use bincode::error::DecodeError;
+use bincode::enc::Encoder;
+use bincode::error::{DecodeError, EncodeError};
 use bincode::{Decode, Encode};
 use core::fmt::Debug;
+use core::marker::PhantomData;
 use cu29::prelude::*;
 use cu29::units::si::f32::Length;
 use cu29::units::si::length::meter;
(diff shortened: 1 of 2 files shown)


Scofield626:
Maybe one thing more to consider here.

I saw the cu_sensor_payloads support for CuImage has this "[with_plane_bytes_mut](https://github.com/copper-project/copper-rs/blob/2e3916c60a9aad562f7637fc2f7aee8db435d9b0/components/payloads/cu_sensor_payloads/src/image.rs#L332)" API, which can use one mutex lock over the whole slice structure. Given `CuHandle::get` calls the `with_inner` every sample here, if I understood correctly, shall we consider similar APIs: 

with_samples<R>(&self, f: impl FnOnce(&[Length], CuDepthMapFormat) -> R) -> R;
with_samples_mut<R>(&mut self, f: impl FnOnce(&mt [Length], CuDepthMapFormat) -> R) -> R;

Imagine a depth map A x B, the lock acquisition naively could have A*B counts? With this BATCHING style, we can provide the flexibility?

components/payloads/cu_sensor_payloads/src/depth.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub stride: u32,
+}
+
+impl CuDepthMapFormat {
+    pub fn is_valid(&self) -> bool {
+        self.width <= self.stride
+    }
+
+    pub fn required_elements(&self) -> usize {
+        self.stride as usize * self.height as usize
```
Scofield626, line 31: can this overflow on u32?

Scofield626 (approved):

components/payloads/cu_sensor_payloads/src/depth.rs
gbin: usize is mostly 64bits nowadays :). If this is on a 32bits system it would mean the addressable memory might be limited to 4GB trying to capture an astronomical scale depthmap of 4GB+. I think it is not worth it to protect against.

gbin:
I am adding a way to have some fixed point elements because most of the hw will generate maps like this and it might not be worth it to make them cconvert their compact formats to floats all the time

PR #1233, opened 2026-07-27
codegen: anytime

Scofield626:
- Generated shape — policy ZST, hoisted job local, base block, k refine blocks.
- Decisions worth reviewing — one slot per node, one SimStep per node, Tov::Range anchoring on end, task_trait_for_specs, the parallel-rt one-stage placement.
- Emitter dedup — the three shared token helpers, and why the fourth emitter forced them.
- Tests — what each compile case pins.

Note:
Parallel-rt continuouly put refine() slot back to back (for the simplicity to make them on the same thread), while sequential one interleaves them. And the replay non-determinism for time-bounded nodes.


gbin:
LGTM

PR #1239, opened 2026-07-31
test: deterministic replay non-regression for anytime nodes

Scofield626:
Adds a non-regression test proving the anytime runner (#1232) did not break deterministic replay.

An anytime node compiles to a base step plus `max_refines` refine steps, and each of those steps reads the clock, stamps status text and (for the base step) freezes the task. That is a lot of new emitted code around the parts replay
depends on, and nothing covered it so far.

`core/cu29_runtime/tests/replay_anytime.rs` builds a small sim app — `src -> planner -> smoother -> sink` — where `planner` is a scored anytime node (`max_age_ms` + `quality_target` + `max_refines`) and `smoother` is a
quality-less one (`max_refines` only), so the test covers the DOA age check, the per-quantum clock reads, an early quality stop, and a stop by plan position.

### DIFF
--- a/core/cu29_runtime/tests/replay_anytime.rs
+++ b/core/cu29_runtime/tests/replay_anytime.rs
@@ -1,0 +1,335 @@
+//! Deterministic-replay non-regression test for foreground anytime nodes.
+//!
+//! An anytime node compiles to a base step plus `max_refines` refine steps, and
+//! each step reads the clock, stamps status text and freezes the task. Replaying
+//! a recorded copperlist must reproduce all of it byte for byte.
+
+#![cfg(all(test, feature = "std"))]
+
+use bincode::{Decode, Encode, config::standard, encode_to_vec};
+use cu29::bincode::de::Decoder;
+use cu29::bincode::enc::Encoder;
+use cu29::bincode::error::{DecodeError, EncodeError};
+use cu29::cutask_anytime::{AnytimeStatus, CuAnytimeTask, Quality, quality_from_f32};
+use cu29::prelude::copper_runtime;
+use cu29::prelude::*;
+use cu29_export::{copperlists_reader, keyframes_reader};
+use cu29_unifiedlog::memmap::{MmapSectionStorage, MmapUnifiedLoggerWrite};
+use cu29_unifiedlog::{UnifiedLogger, UnifiedLoggerBuilder, UnifiedLoggerIOReader};
+use std::fs;
+use std::path::Path;
+use std::sync::{Arc, Mutex};
+
+/// What the sink saw, one entry per copperlist: the payload and the status text
+/// the anytime runner stamped on it.
+static OBSERVED: Mutex<Vec<(u32, String)>> = Mutex::new(Vec::new());
+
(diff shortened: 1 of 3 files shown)


core/cu29_runtime/tests/replay_anytime.rs
```diff
@@ -0,0 +1,335 @@
...
+    // Same input, fresh application: the base/refine steps must produce the same
+    // payloads, the same status stamps and the same timings.
+    let (second_cl, second_kf, second_observed) = record_reference_run(&second_path)?;
+    assert_eq!(second_observed, first_observed);
+    assert_eq!(encode_bytes(&second_cl), encode_bytes(&first_cl));
+    assert_eq!(encode_bytes(&second_kf), encode_bytes(&first_kf));
+
+    // Replaying the recorded copperlist reproduces it byte for byte, keyframe
+    // included: one anytime node is still one replay step and one freeze.
+    let (replayed_cl, replayed_kf) = replay_run(&replay_path, &first_cl, &first_kf)?;
```
gbin: This is only one copperlist that is equal to the keyframe so basically there is no test in drift.

core/cu29_runtime/tests/replay_anytime_config.ron
```diff
@@ -0,0 +1,52 @@
...
+    tasks: [
+        (
+            id: "src",
+            type: "TargetSrc",
+            run_in_sim: true,
+        ),
+        (
+            id: "planner",
+            type: "IncrementalPlanner",
+            run_in_sim: true,
```
gbin, line 16: so ok here you ask copper to run it in sim.

core/cu29_runtime/tests/replay_anytime.rs
```diff
@@ -0,0 +1,335 @@
...
+    let mut noop = |_step: default::SimStep<'_>| SimOverride::ExecuteByRuntime;
+
+    let mut app = AnytimeReplayApp::builder()
+        .with_clock(clock)
+        .with_logger::<MmapSectionStorage, MmapUnifiedLoggerWrite>(logger)
+        .with_sim_callback(&mut noop)
+        .build()?;
+
+    app.start_all_tasks(&mut noop)?;
+    app.replay_recorded_copperlist(&clock_mock, recorded_cl, Some(keyframe))?;
```
gbin: this API is super confusing but here you did not replay anything actually lol.

it should be more like:
let mut callback = |step: default::SimStep<'_>| {
      default::recorded_debug_replay_step(step, recorded_cl)
  };

  app.run_one_iteration(&mut callback)?;

core/cu29_runtime/tests/replay_anytime.rs
gbin: Check for https://github.com/copper-project/copper-rs/blob/master/examples/cu_caterpillar/src/determinism.rs#L94

For an actual resim / deterministic raplay test. I tweak a bunch for times in there to be sure they are the only drifts in there (I compare the output in binary afterwards).

gbin (changes requested):
this is a noop test in terms of determinism.

gbin (approved):

PR #1242, opened 2026-07-31
Graphical representation for the planner

gbin:
Implements "just plan" building an svg explaining the planned scheduling from copper.

## Related issues
- Closes #

## Reminder


support/just/plan.just
```diff
@@ -0,0 +1,58 @@
+# Shared schedule visualizer used by the workspace and example-local Justfiles.
+PLAN_WORKSPACE_ROOT := shell('git -C "$1" rev-parse --show-toplevel', justfile_directory())
+
+# Render the generated schedule for a config in the invocation directory.
+plan mission="" features="" config="":
```
Scofield626: this might be problematic:

`just plan features=mock` sets mission_value="features=mock" and passes --mission features=mock. Shall we have a variadic key=value parser?

support/just/plan.just
```diff
@@ -0,0 +1,58 @@
+# Shared schedule visualizer used by the workspace and example-local Justfiles.
+PLAN_WORKSPACE_ROOT := shell('git -C "$1" rev-parse --show-toplevel', justfile_directory())
```
Scofield626: do we really want to do rev-parse, I think this is the key reason that breaks the CI

Scofield626 (approved):

PR #1247, opened 2026-08-03
cu-rrt-star: anytime RRT* planner component, with a closed-loop example

Scofield626:
An RRT* planner (Karaman and Frazzoli 2011, with the branch-and-bound pruning of the anytime RRT* from "Anytime ROS2: Timely Task Completion in Non-preemptible Robotic Systems", RTAS 2026) as the first real algorithm on `CuAnytimeTask`.

- `components/tasks/cu_rrt_star` — the component. `base()` grows a tree until a first path; each `refine()` runs one block of iterations and republishes only a shorter path. Every published path is drivable as published. Quality is straight-line distance over path cost, so a `quality_target` is portable between maps. Remote debug reads a small projection, not the tree.
- `components/tasks/cu_rrt_star/tests` — one application running the same planner under a quick (`max_refines: 2`, `quality_target: 0.93`) and a thorough (`max_refines: 24`, `max_stall: 4`) policy on the same seed, so the thorough path can never be longer.
- `examples/cu_anytime_rrt_star` — closed-loop navigation: a simulated robot patrols the depot map, replans from its live pose every copperlist, drawn in Rerun.

11 tests cover the planner geometry, determinism, the anytime contract, and the dual-policy application run.


gbin:
Can we have a proper copper component for it? and add an example using it?

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+}
+
+/// A round obstacle: the planner rejects any point or segment within `radius`
+/// of `center`.
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+)]
+pub struct Obstacle {
+    pub center: Point2,
+    pub radius: f32,
```
gbin: needs to be cu29-units.

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+/// [`crate::PlanRequest`] without allocating; must stay at or under 32 for
+/// the same serde reason as [`MAX_WAYPOINTS`].
+pub const MAX_OBSTACLES: usize = 16;
+
+/// A point of the planar world, in meters.
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+)]
+pub struct Point2 {
+    pub x: f32,
```
gbin: needs to be units

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+/// Obstacles carried by a [`World`]. Bounded so the map can travel inside a
+/// [`crate::PlanRequest`] without allocating; must stay at or under 32 for
+/// the same serde reason as [`MAX_WAYPOINTS`].
+pub const MAX_OBSTACLES: usize = 16;
+
+/// A point of the planar world, in meters.
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+)]
+pub struct Point2 {
```
gbin: it cannot be unified with what we have as a standard struct?

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+)]
+pub struct Obstacle {
+    pub center: Point2,
+    pub radius: f32,
+}
+
+impl Obstacle {
+    pub const fn new(center: Point2, radius: f32) -> Self {
```
gbin: is the algorithm mandate round obstacles?

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+}
+
+/// The rectangular world `0..width` x `0..height` and its obstacles.
+///
+/// The obstacle storage is a fixed array so the map can travel inside a
+/// [`crate::PlanRequest`]: the planner has no map of its own, every job
+/// carries the one it must solve.
+#[derive(Default, Debug, Clone, Encode, Decode, Serialize, Deserialize, Reflect)]
+pub struct World {
+    pub width: f32,
```
gbin: units

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+    pub height: f32,
+    pub obstacles: [Obstacle; MAX_OBSTACLES],
```
gbin, line 200: should we start a notion of world in copper with a local transform? this is very narrow 16 spheres in a rectangle, also it copies the world.
Maybe a view on a CuHandle is the way to do it? imagine a scene in a self driving, it can be way more trickier than this.

components/tasks/cu_rrt_star/src/rrt.rs
```diff
@@ -0,0 +1,833 @@
...
+
+/// xorshift64*, so a given seed always replays the same tree.
+#[derive(Debug, Clone, Reflect)]
+pub struct Rng(u64);
+
+impl Rng {
+    pub fn new(seed: u64) -> Self {
+        // splitmix64 finalizer: consecutive seeds must not start on neighboring
+        // states, otherwise consecutive jobs explore almost the same tree.
+        let mut state = seed.wrapping_add(0x9E37_79B9_7F4A_7C15);
```
gbin: what are we doing reimplementing this here?!

components/tasks/cu_rrt_star/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+
+/// One planning problem. The map travels with the job, so the source owns it
+/// and may change it between jobs. A different seed per job makes each job a
+/// fresh RRT* run rather than a replay of the previous one.
+#[derive(Default, Debug, Clone, Encode, Decode, Serialize, Deserialize, Reflect)]
+pub struct PlanRequest {
+    pub world: World,
+    pub start: Point2,
+    pub goal: Point2,
+    pub seed: u64,
```
gbin: what is going on here? even with the explanation for the love of my life I don't understand why any random seed would land in a live message.

components/tasks/cu_rrt_star/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+/// not the thousands of tree nodes behind it.
+///
+/// The fields are read through `Reflect`, which the compiler cannot see.
+#[allow(dead_code)]
+#[derive(Default, Debug, Reflect)]
+pub struct PlannerDebugState {
+    pub iterations: u32,
+    pub tree_size: u32,
+    /// Nodes on the best path before it is shortcut for publication.
+    pub tree_path_len: u32,
```
gbin, line 72: units

components/tasks/cu_rrt_star/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// the best result so far, so the runtime can publish it at any stop point.
+    fn publish(&mut self, output: &mut CuMsg<PlanPath>) -> Quality {
+        if let Some(planner) = self.planner.as_ref()
+            && planner.has_solution()
+            && planner.best_cost() < self.published_cost
+        {
+            let mut waypoints = [Point2::default(); MAX_WAYPOINTS];
+            // A path too long to represent is not published: the output keeps
+            // the last valid one and a later quantum tries again.
+            if let Some(len) = planner.write_path(&mut waypoints) {
```
gbin, line 144: maybe: update the next output in place and just copy it last minute instead of a custom copy here ? (might save some copy time, do it only if it does)

components/tasks/cu_rrt_star/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub iterations: u32,
+}
+
+/// What a remote debugger sees of a planner node: the progress of the job,
+/// not the thousands of tree nodes behind it.
+///
+/// The fields are read through `Reflect`, which the compiler cannot see.
+#[allow(dead_code)]
+#[derive(Default, Debug, Reflect)]
+pub struct PlannerDebugState {
```
gbin, line 68: is this for only when the debug feature is active?

components/tasks/cu_rrt_star/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+An RRT* planner (Karaman and Frazzoli) packaged as a Copper anytime task.
+`base()` grows the tree until it has a first path and publishes it; every
+`refine()` runs one more block of iterations and republishes only when the
+path got shorter. The RON `anytime:` policy — not the task — decides how many
+refinement quanta run.
+
+### Input and output
+
+- Input: `cu_rrt_star::PlanRequest` — the world (up to `MAX_OBSTACLES` (16)
```
gbin, line 11: templatign it to make it 2d and 3d would be difficult?

gbin (changes requested):

gbin (approved):

PR #1248, opened 2026-08-03
feat: background placement for anytime tasks

Scofield626:
An `anytime:` node can now set `background: true`. Rather than new concurrent machinery, it reuses `CuAsyncTask`: the node compiles to `CuAsyncTask<CuAnytimeRunner<T, Policy>, O>`, keeps its single poll step in the plan, and the worker runs one whole job per dispatch — age check, `base()`, refine quanta under the policy, `finish`.

- `CuAnytimeRunner` (doc(hidden)) is a plain `CuTask`, so everything `CuAsyncTask` provides applies unchanged: latest-input dispatch, buffered output, `ready_at` replay gating, errors surfacing at the next poll, no freeze mid-job.
- The runner drives the documented per-job lifecycle on the worker: `preprocess` → `base` → `refine` × N → `postprocess`.
- Every configured policy bound (quality target, time budget, age, stall) is checked between quanta, exactly as in the emitted foreground blocks; unset knobs const-fold away. `MAX_REFINES` is the one const the foreground plan never reads (it encodes the count positionally), and `max_refines` is optional here — a time bound alone closes the loop.
- Codegen resolves the wrapped type once (`CuTaskSpecSet::inner_task_types`); background anytime nodes keep their single plan step, so keyframes, sim, and parallel-rt see an ordinary backgrounded task.

Tests: runner unit tests for every stop cause and for hook ordering, a composition test through `CuAsyncTask`, compile-pass apps in default and sim mode, and an example running foreground and background nodes off one task type.


gbin (changes requested):
Edit: ok ok sorry there might have been just a readability problem with the variables with "parsed" in the name

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -436,6 +442,20 @@ fn stamp<O: CuMsgPayload>(
...
+/// within the age limit.
+#[doc(hidden)]
+#[inline(always)]
+pub fn anchor_from_tov(tov: Tov, now: CuTime) -> CuTime {
+    match tov {
+        Tov::Time(time) => time,
+        Tov::Range(range) => range.start,
+        Tov::None => now,
+    }
+}
```
gbin, line 457: - Background anytime: there is currently an unconditional clock read because of this introduction:

  let start = ctx.now();
  let anchor = anchor_from_tov(input.tov, start);

  So if neither TIME_BUDGET nor MAX_AGE is configured, ctx.now() is unnecessary. The resulting elapsed value is not used by run_job, and the
  foreground path already avoids this cost.

  I would align the background runner with the foreground behavior:

  let start = if P::TIME_BUDGET.is_some() || P::MAX_AGE.is_some() {
      ctx.now()
  } else {
      CuTime::default()
  };

  let anchor = if P::MAX_AGE.is_some() {
      anchor_from_tov(input.tov, start)
  } else {
      start
  };

core/cu29_runtime/src/cutask_anytime.rs
```diff
@@ -469,12 +489,213 @@ pub fn abort_at_base<O: CuMsgPayload>(
...
+/// day forwards per-cycle hooks cannot double-call the task.
+#[doc(hidden)]
+#[derive(Reflect)]
+#[reflect(no_field_bounds, from_reflect = false, type_path = false)]
+pub struct CuAnytimeRunner<T, P>
+where
+    T: Send + Sync + 'static,
+    P: Send + Sync + 'static,
+{
+    #[reflect(ignore)]
```
gbin, line 513: does this mean it becomes invisible to the debug API?

gbin (approved):
We can probably optimize this more, the indirection adds jumps in memory etc... but as the external interface is ok, we can always do that in a second pass. Let's land this.

Issue #1230, opened 2026-07-26
feature: Sensor parameters / calibration support


gbin:
## Problem statement

When we have a CuDepthMap or even a simple image, we might need down the pipeline a way to know the original physical characteristics of the sensor like the calibration and field of view. 

Today we have a way to override parameters on the original source for each specific instances of a robot but we have no way to share that info with a task downstream. 

## Proposed solution

Maybe something that looks like the resources instead of repeating it on the copperlist.

## Impact / scope (optional)
- Affected crates, platforms, or features:

PR #1253, opened 2026-08-04
feat(payloads): add typed dynamic camera calibration model

Daniiiil1:
## Summary

Adds a standard, fixed-size camera model for image and depth-map pipelines, framed specifically as dynamic calibration. The distortion family is selected at compile time with `CuCameraModel<D>`; a running robot can update intrinsics and coefficients, but cannot change the mathematical distortion model at runtime.

The propagation path uses the existing latched-state primitive: a source emits the full calibration only when it appears or changes, consumers cache it locally, and later cycles carry `NoChange`. This keeps dynamic calibration deterministic under unified logging and replay without introducing mutable global state on the realtime path.

## Related issues

- Relates to #1230

## Changes

- Add `CuCameraIntrinsics` with an explicit pixel-center convention, validated pinhole parameters, FOV calculations, and rectified pixel-ray conversion including skew.
- Add compile-time distortion types for no distortion, plumb-bob, rational-polynomial, and equidistant models.
- Encode each coefficient count in the constructor type, so invalid counts fail to compile rather than being rejected at runtime.
- Add generic `CuCameraModel<D>`, `CuCameraModelUpdate<D>`, and `CuCameraModelState<D>` for typed dynamic-calibration propagation.
- Support cropped cameras whose principal point lies outside the delivered image.
- Keep the model available under `no_std`; `libm` provides portable atan calculations.
- Document the dedicated-output propagation pattern.

## Verification

- `cargo test -p cu-sensor-payloads`: 43 unit tests, integration tests, and doc tests pass, including a compile-fail coefficient-count test.
- `cargo check -p cu-sensor-payloads --no-default-features`: pass.
- `cargo check -p cu-sensor-payloads --features reflect`: pass.
- `cargo clippy -p cu-sensor-payloads --all-targets -- -D warnings`: pass.
- `cargo clippy -p cu-sensor-payloads --no-default-features --lib -- -D warnings`: pass.
- `cargo fmt --all -- --check` and `git diff --check`: pass.

## Design questions

This remains a draft because issue #1230 raises a topology decision. I chose a dedicated latched output over runtime-global state so calibration changes remain visible to record/replay. Maintainer feedback would be especially useful on whether Copper should keep this contract or eventually generate a typed graph-level side-data channel.

## Local environment note

The root `justfile` could not be evaluated from this checkout because the absolute path contains a space and its `ROOT` shell argument is unquoted. The focused CI-equivalent checks above all passed.

### DIFF
--- a/components/payloads/cu_sensor_payloads/Cargo.toml
+++ b/components/payloads/cu_sensor_payloads/Cargo.toml
@@ -25,6 +25,7 @@ cu29-clock = { path = "../../../core/cu29_clock", version = "1.1.0-dev", default
 cu29 = { path = "../../../core/cu29", version = "1.1.0-dev", default-features = false, features = [
   "units",
 ] }
+libm = { version = "0.2", default-features = false }
 rerun = { workspace = true, optional = true }
 image = { workspace = true, default-features = false, optional = true }
 kornia-image = { version = "0.1.10", optional = true }
--- a/components/payloads/cu_sensor_payloads/README.md
+++ b/components/payloads/cu_sensor_payloads/README.md
@@ -3,7 +3,20 @@
 Standardized sensor payload definitions for Copper.
 
 The crate contains common payload types used by Copper sources, tasks, and
-sinks, including image and point-cloud related data structures.
+sinks, including image, depth-map, camera-calibration, and point-cloud data
+structures.
+
+## Camera calibration propagation
+
+`CuCameraModel` provides a fixed-size, allocation-free camera model containing
+pinhole intrinsics and a standard distortion model. It is intentionally separate
+from `CuImage` and `CuDepthMap`: a source can expose the model on a dedicated
+output as `CuCameraModelUpdate` and send `Set(model)` only when calibration first
+becomes available or changes. Later cycles carry `NoChange`, while each consumer
+keeps a `CuCameraModelState` cache.
+
+This keeps the per-frame payload small without hiding calibration changes from
+Copper's unified log and deterministic replay.
 
 ## Features
 
@@ -12,4 +25,3 @@ sinks, including image and point-cloud related data structures.
 - `image`
 - `kornia`
 - `rerun`
-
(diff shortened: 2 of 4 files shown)


gbin:
Thanks for your contribution. We definitely need to standardize calibration in the stack.

Note: today in the examples etc we only have some form of static calibration (in the RON files), and the pattern we encourage our users to use is to use the composition of ron files to make the configuration of any *instance* of a robot specialized with a static calibration: ie robot #3, get that calibration, robot #4, another one etc...

Your PR adds some kind of dynamic calibration to the system, which does exist in some kind of robots, one classic example are Tesla cars when you drive around and the calibration is picked up from known straight lines on the road.

So maybe first step would be to reframe that in that context: ie. dynamic calibration.

Then ... why is the TYPE of calibration dynamic? this, to be clear, will *never* happen in real life, you will never have a robot suddenly deciding to change the type of calibration for a camera live (maybe in some kind of lab obscure condition or research but this would be our of scope).

The second fix would be to hardcode the type of calibration as a type parameter in for the rust struct.

Hopefully it makes sense.

Thanks

Daniiiil1:
Thanks, that distinction makes sense. I’ve reframed the feature as dynamic calibration and moved the distortion family into the Rust type: `CuCameraModel<D>`. The robot may update intrinsics or coefficients, but changing from, say, plumb-bob to equidistant now requires a different compiled type. Each model also takes a fixed-size coefficient array, so the wrong count is a compile-time error. The package tests, no_std check, reflect check, and clippy all pass on the updated branch.

gbin:
@Daniiiil1 any actual robots you are running this on?

Daniiiil1:
Not this exact Copper integration yet, so I don’t want to overstate the level of hardware validation.

I have worked with real STM32/ARM-based control systems and robotic hardware, including an autonomous hydroponic greenhouse with mobile robots, sensors, actuators, and closed-loop control. However, this particular camera-calibration implementation has currently been validated in software through the typed update path, tests, logging/replay behavior, and `no_std` builds.

If you can point me toward the hardware or reference robot normally used by the Copper team, I’d be happy to adapt the example and validate the full calibration-update path on it.

gbin:
This is just an LLM spamming for you to find a job so I am not super thrilled about merging this sorry.

PR #1254, opened 2026-08-05
cu-spatial-payloads: unit-typed points, BBox, Transform3D point ops

Scofield626:
Shared geometry vocabulary, split out of #1247 as requested in review (https://github.com/copper-project/copper-rs/pull/1247#discussion_r3713888734): planners and future world views should compose over one standard point, not re-invent it per component.

- `Point2<L>`/`Point3<L>` generic over the coordinate type, with `Point2f/2d/3f/3d` aliases over the f32/f64 `Length` wrappers and `Point2u/2i` pixel aliases for image-space geometry; `from_meters`, `distance`, `lerp`, `xy`/`with_z`.
- `BBox<P>` axis-aligned bounding boxes (`BBox2f/2d/3f/3d`, plus `BBox2u/2i` for images), `contains`.
- `Transform3D::transform_point` / `transform_vector` / `position`, implemented on the native backend path (glam affine when the feature is on, raw matrix otherwise).
- `no_std` kept; `sqrt` comes from `libm`, already used by `cu29-runtime` and `cu_peer_triangulation`.

Addressed from review:
- `Aabb` renamed to `BBox`.
- Integer coordinates: the structs are generic, `Point2u/2i` + `BBox2u/2i` aliases added for images.
- `Clearance` dropped from this crate as too specific; it moves into `cu_rrt_star` (#1247).
- SoA: scalar type docs now point large batches at an SoA layout. The SoA point/box types come in a small PR stacked on this one — `cu29_soa_derive` first needs to learn generic structs like `Point2<L>`, then bulk kernels (distance, lerp, box ops, `transform_points`) with AVX2/AVX512 codegen checked.

First consumer: #1247 switches `cu_rrt_star` to `Point2f`.

Noted while implementing, left out of scope: `Transform3D::translation()`/`rotation()` read row-major cells, but the glam backend (default feature) stores column-major, so they return zeros/transposed values when glam is enabled. The new ops avoid `to_matrix()` for that reason. Happy to file/fix separately.

### DIFF
--- a/components/payloads/cu_spatial_payloads/src/geometry.rs
+++ b/components/payloads/cu_spatial_payloads/src/geometry.rs
@@ -4,18 +4,23 @@
 use bincode::{Decode, Encode};
 use core::fmt::Debug;
 use cu29::prelude::*;
+use cu29::units::si::area::square_meter;
+use cu29::units::si::f32::Area as Area32;
 use cu29::units::si::f32::Length as Length32;
+use cu29::units::si::f64::Area as Area64;
 use cu29::units::si::f64::Length as Length64;
 use cu29::units::si::length::meter;
+use cu29_soa_derive::Soa;
 use serde::{Deserialize, Serialize};
 
 /// A 2D point with unit-typed coordinates.
 ///
-/// This is a scalar type; for large batches consider an SoA layout so bulk
-/// operations vectorize (see `cu29_soa_derive`).
+/// This is a scalar type; for large batches use [`Point2Soa`] so bulk
+/// operations vectorize.
 #[derive(
-    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect, Soa,
 )]
+#[reflect(from_reflect = false)]
 pub struct Point2<L: Copy + Debug + 'static> {
     pub x: L,
     pub y: L,
@@ -23,11 +28,12 @@ pub struct Point2<L: Copy + Debug + 'static> {
 
 /// A 3D point with unit-typed coordinates.
 ///
-/// This is a scalar type; for large batches consider an SoA layout so bulk
-/// operations vectorize (see `cu29_soa_derive`).
+/// This is a scalar type; for large batches use [`Point3Soa`] so bulk
+/// operations vectorize.
 #[derive(
-    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect, Soa,
 )]
+#[reflect(from_reflect = false)]
 pub struct Point3<L: Copy + Debug + 'static> {
     pub x: L,
     pub y: L,
@@ -43,6 +49,13 @@ pub type Point3d = Point3<Length64>;
 pub type Point2u = Point2<u32>;
 /// A 2D point in signed pixel coordinates, for image-space offsets.
 pub type Point2i = Point2<i32>;
+
+pub type Point2fSoa<const N: usize> = Point2Soa<Length32, N>;
+pub type Point2dSoa<const N: usize> = Point2Soa<Length64, N>;
+pub type Point3fSoa<const N: usize> = Point3Soa<Length32, N>;
+pub type Point3dSoa<const N: usize> = Point3Soa<Length64, N>;
+pub type Point2uSoa<const N: usize> = Point2Soa<u32, N>;
+pub type Point2iSoa<const N: usize> = Point2Soa<i32, N>;
 
 impl<L: Copy + Debug + 'static> Point2<L> {
     pub const fn new(x: L, y: L) -> Self {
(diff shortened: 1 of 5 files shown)


components/payloads/cu_spatial_payloads/src/geometry.rs
```diff
@@ -0,0 +1,216 @@
...
+pub struct Point2<L: Copy + Debug + 'static> {
+    pub x: L,
+    pub y: L,
+}
+
+/// A 3D point with unit-typed coordinates.
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+)]
+pub struct Point3<L: Copy + Debug + 'static> {
```
gbin, line 37: we immediately need SOA versions of this otherwise users will write very inefficient bulk version of embedded algorithms

components/payloads/cu_spatial_payloads/src/geometry.rs
```diff
@@ -0,0 +1,216 @@
...
+macro_rules! impl_point_metrics {
+    ($len:ty, $scalar:ty, $sqrt:path) => {
+        impl Point2<$len> {
+            pub fn from_meters(x: $scalar, y: $scalar) -> Self {
+                Self::new(<$len>::new::<meter>(x), <$len>::new::<meter>(y))
+            }
+
+            /// Euclidean distance to `other`.
+            pub fn distance(self, other: Self) -> $len {
+                let (dx, dy) = ((self.x - other.x).raw(), (self.y - other.y).raw());
```
gbin, line 91: the SOA version will be 4x to 16x faster computing that using SIMD

components/payloads/cu_spatial_payloads/src/geometry.rs
```diff
@@ -0,0 +1,216 @@
...
+                <$len>::new::<meter>($sqrt(dx * dx + dy * dy))
+            }
+
+            /// The point at `ratio` of the way toward `other`: 0 is `self`,
+            /// 1 is `other`.
+            pub fn lerp(self, other: Self, ratio: $scalar) -> Self {
```
gbin, line 97: same for SOA, point in the comment that if you have a large vector of those to consider an SOA

components/payloads/cu_spatial_payloads/src/geometry.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+impl_point_metrics!(Length32, f32, libm::sqrtf);
+impl_point_metrics!(Length64, f64, libm::sqrt);
+
+/// An axis-aligned box; the point type carries the dimension.
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Encode, Decode, Serialize, Deserialize, Reflect,
+)]
+pub struct Aabb<P: Copy + Debug + 'static> {
```
gbin: bounding boxes comes in bulk too: imagine all the anchors in an ML bases CV algo like Yolo.

We absolutely need to think about memory layout / SOA. If we force in all scalars like this it might be a performance disaster.

components/payloads/cu_spatial_payloads/src/geometry.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub min: P,
+    pub max: P,
+}
+
+pub type Aabb2f = Aabb<Point2f>;
+pub type Aabb2d = Aabb<Point2d>;
+pub type Aabb3f = Aabb<Point3f>;
+pub type Aabb3d = Aabb<Point3d>;
```
gbin: we need ints right away for images.

components/payloads/cu_spatial_payloads/src/geometry.rs
gbin: for bulk operations like intersections of bboxes, union etc...

gbin: maybe BBox or BoundingBox?

components/payloads/cu_spatial_payloads/src/geometry.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+/// Clearance queries against the occupied geometry of a space.
+///
+/// The clearance is a signed distance: positive in free space, zero on a
+/// surface, negative inside an obstacle or out of bounds. What is free is the
+/// caller's predicate - `clearance(p) > robot_radius` - so one space serves
+/// robots of any size.
+pub trait Clearance {
```
gbin: That is too specific, we are opening a can of worms here.
Like clearance of a path, points in polygons, unions etc.. maybe for a future pass on this lib.

components/payloads/cu_spatial_payloads/src/lib.rs
```diff
@@ -267,6 +273,107 @@ impl<T: Copy + Debug + Default + 'static> TransformInner<T> {
 impl_transform_accessors!(f32, Length32, Angle32);
 impl_transform_accessors!(f64, Length64, Angle64);
 
+macro_rules! impl_transform_points {
+    ($ty:ty, $len:ty, $variant:ident, $vec:ty) => {
+        impl Transform3D<$ty> {
+            /// The translation as a point: where the pose sits in its parent
+            /// frame.
+            pub fn position(&self) -> Point3<$len> {
```
gbin, line 282: here too we need to provide mass vectors or array optimized version of that we check runs full speed on AVX2 and AVX512 minimum.

gbin (changes requested):

gbin (approved):

PR #1260, opened 2026-08-06
cu29-units: add curated const construction

gbin:
Add const construction directly to the existing Copper unit wrapper types, without a new crate or new Length/Angle types.\n\nThe initial, deliberately limited robotics motion surface covers:\n\n- Length: meter, centimeter, millimeter\n- Angle: radian, degree\n- Time: second, millisecond, microsecond, nanosecond\n- Velocity: meter per second, kilometer per hour\n- AngularVelocity: radian per second, degree per second\n- Acceleration: meter per second squared, standard gravity\n\nSupported wrappers gain new_const::<U>(); raw() is const. Hidden sealed per-dimension registries reject unsupported units at compile time, while the existing runtime new::<U>() remains comprehensive. Conversion tests match uom exactly for f32/f64 and std/no_std.\n\nVerification:\n- cargo test -p cu29-units\n- cargo test -p cu29-units --no-default-features\n- cargo clippy -p cu29-units --all-targets -- -D warnings\n- cargo clippy -p cu29-units --all-targets --no-default-features -- -D warnings\n- std and no_std doctests

### DIFF
--- a/core/cu29_units/README.md
+++ b/core/cu29_units/README.md
@@ -24,3 +24,29 @@ You can also depend on `cu29-units` directly:
 use cu29_units::si::f32::Length;
 use cu29_units::si::length::meter;
 ```
+
+## Const construction
+
+The existing wrapper types provide const construction for a deliberately limited set of
+robotics motion units:
+
+```rust
+use cu29_units::si::angle::degree;
+use cu29_units::si::f32::{Angle, Length};
+use cu29_units::si::length::meter;
+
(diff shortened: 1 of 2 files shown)


gbin:
This is not strictly necessary, only if we want to combine unit conversion AND constness. 
Which looks like not such a big deal. I leave it here as a draft for now.

gbin:
this is not strictly necessary with {value: ...} contruction directly with the pure SI unit so let's not complicate that if it is not necessary

PR #1272, opened 2026-08-14
feat(runtime): typestate lifecycle wrapper for CuApplication

### LINKED ISSUE #1091: feature: User friendly way to notify a lifecycle violation in the runtime
## Problem statement

Today we have several levels of APIs for the runtime that allows a full control from the user of the framework over the lifecycle of it.

The issue is that the user can put the stack in a bad state without noticing: for example calling start_all_tasks twice or mixing it with application.run

## Proposed solution

Add a small enum based state machine to track the validity of those lifecycle changes and emit an error accordingly. 

## Alternatives considered (optional)

Make that debug only but this should be so small it might not worth it

## Impact / scope (optional)
- Affected crates, platforms, or features:

arunvenkatadri:
## Summary

Adds a compile-time enforced lifecycle for Copper applications, following the typestate design discussed in #1091, and makes it the **default construction path**: the generated builder ends in `build_app()`, which returns the application already wrapped in its lifecycle typestate. Each state is a distinct type that only exposes the transitions that are legal from it, so mistakes like starting an already running application, iterating before start, or reusing a stale handle fail to compile instead of misbehaving at runtime.

## Related issues
- Closes #1091
- Supersedes #1275 (examples port, folded in here)

## Changes
- `CuAppLifecycle<S, L, A, State>` in `cu29_runtime::app` (plus `CuStdAppLifecycle` alias), with states `Initialized`, `Running`, `Stopped`, `Faulted` and sealed `Startable`/`Stoppable` traits carrying `#[diagnostic::on_unimplemented]` hints
- New `CuSimAppLifecycle` counterpart for `CuSimApplication`: same states, each transition threads the `sim_callback`
- The generated application builder gains **`build_app()`** returning the wrapped app; `build()` and the generated inherent lifecycle convenience methods are `#[deprecated]`. The `CuApplication`/`CuSimApplication` traits are unchanged so framework-level harnesses keep working
- Transitions consume the wrapper and return it typed with the new state; a failed transition hands the application back typed `Faulted` inside `LifecycleError`/`SimLifecycleError` for cleanup, and `From<...> for CuError` keeps `?` working. Restarting from `Stopped` is legal (mission chaining)
- `inner_mut()` is exposed only in the `Initialized` state for pre-start configuration; started states are read-only so transitions cannot be bypassed mid-flight
- Whole tree migrated to `build_app()`: examples (sim included — balancebot sim driver and caterpillar determinism are on `CuSimAppLifecycle`), component testers, core tests, and the cargo-cunew templates. Replay harnesses that need mid-flight runtime access (keyframe locking) stay on the raw trait API behind a commented `allow(deprecated)` or `into_inner()`
- api/v1 snapshots regenerated

The migration surfaced real latent misuse now caught at compile time: six examples double-started/stopped tasks around `run()`, and five hardware testers iterated without ever starting their tasks.

Verification:
- Unit tests in `cu29_runtime::app` for both wrappers: full cycle, restart, failed start/stop recovery through `Faulted`, `?` propagation, sim callback threading
- trybuild compile-fail cases (start on a `Running` app, reuse of a consumed handle) and a compile-pass case covering the legal flows
- Clippy `-D warnings` clean across all migrated crates; `no_std` verified on `thumbv7em-none-eabihf`; both embedded example crates build for their targets

## Reminder
- [x] I ran `just` from the repo root

## Additional context (Not AI Generated)
- Of the construction-flow options considered, I picked the one with the smallest macro delta: a new terminal on the existing builder rather than a parallel deprecated builder
- A resim-aware typestate for the replay engines is a possible follow-up if wanted
- Task-level lifecycles are untouched: the runtime already sequences those internally
- Local verification notes (macOS): fmt, typos, api-check and the test suites for `cu29-runtime` and `cu29-derive` ran clean; the python/gstreamer/glib-linked targets could not link on this machine, relying on CI for those
- Note this is my first issue so any guidance here is helpful. I'm hoping to contribute to this project more.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -387,7 +387,7 @@ simplelog = "0.12"
 socket2 = "0.6"
 stm32h7xx-hal = { version = "0.16.0", default-features = false }
 strfmt = "0.2"
-zenoh = { version = "1.7.0", default-features = false }
+zenoh = { version = "~1.10.0", default-features = false }
 
 [workspace.metadata.cargo-shear]
 ignored = ["cu-consolemon"]
--- a/api/v1/cu29-runtime.txt
+++ b/api/v1/cu29-runtime.txt
@@ -2,10 +2,10 @@ pub mod cu29_runtime
 pub mod cu29_runtime::app
 pub struct cu29_runtime::app::CuAppLifecycle<S, L, A, State>
 impl<S, L, A, State> cu29_runtime::app::CuAppLifecycle<S, L, A, State> where S: cu29_unifiedlog::SectionStorage, L: cu29_unifiedlog::UnifiedLogWrite<S> + 'static, A: cu29_runtime::app::CuApplication<S, L>, State: cu29_runtime::app::Startable
-pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::run(self) -> cu29_runtime::app::TransitionResult<S, L, A, cu29_runtime::app::Stopped>
-pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::start_all_tasks(self) -> cu29_runtime::app::TransitionResult<S, L, A, cu29_runtime::app::Running>
+pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::run_until_shutdown(self) -> cu29_runtime::app::TransitionResult<S, L, A, cu29_runtime::app::Stopped>
+pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::start(self) -> cu29_runtime::app::TransitionResult<S, L, A, cu29_runtime::app::Running>
 impl<S, L, A, State> cu29_runtime::app::CuAppLifecycle<S, L, A, State> where S: cu29_unifiedlog::SectionStorage, L: cu29_unifiedlog::UnifiedLogWrite<S> + 'static, A: cu29_runtime::app::CuApplication<S, L>, State: cu29_runtime::app::Stoppable
-pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::stop_all_tasks(self) -> cu29_runtime::app::TransitionResult<S, L, A, cu29_runtime::app::Stopped>
+pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::stop(self) -> cu29_runtime::app::TransitionResult<S, L, A, cu29_runtime::app::Stopped>
 impl<S, L, A, State> cu29_runtime::app::CuAppLifecycle<S, L, A, State>
 pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::inner(&self) -> &A
 pub fn cu29_runtime::app::CuAppLifecycle<S, L, A, State>::into_inner(self) -> A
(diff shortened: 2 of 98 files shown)


gbin:
Thanks @arunvenkatadri can you add another PR in a PR chain on top of this one porting all the examples to the new API so we can check how it looks like on the user side?

gbin:
urgh closed by mistake

arunvenkatadri:
Examples port is up in #1275 (stacked on this one — only the top commit is new there). Porting surfaced six examples that were double-starting/stopping tasks by calling start_all_tasks/stop_all_tasks around run(); the typestate rejects that at compile time, so they now use a single run() transition. Ergonomics notes and the non-portable cases (CuSimApplication takes a sim_callback per call, so sim/resim examples are out of scope) are written up in that PR's description.

arunvenkatadri:
Thanks for the review @gbin — note written by me, not AI generated: I'm still learning onboard software, so I really appreciated this feedback... learning from the best :) On the construction flow, of the options considered I went with the one that resulted in the smallest macro delta: same builder, new `build_app()` terminal returning the lifecycle handle, with `build()` and the inherent lifecycle methods deprecated — rather than a whole parallel builder.

Reworked per both objections:

**1. Insertion point — the typestate is now the default path.** The generated builder ends in `build_app()`, which hands out the application already wrapped (`CuAppLifecycle`, or `CuSimAppLifecycle` in sim mode). `build()` and the generated inherent `start_all_tasks()`/`run()`/... convenience methods carry `#[deprecated]` pointing at it; the `CuApplication`/`CuSimApplication` traits are unchanged so framework-level code keeps working. The user side is now *less* than before, not more — this is the cargo-cunew template verbatim:

```rust
let application = MyApplication::builder()
    .with_log_path(logger_path, SLAB_SIZE)
    .expect("Failed to setup logger.")
    .build_app()
    .expect("Failed to create application.");
let stopped = application.run().expect("Failed to run application.");
```

**2. The two robots are ported**, sim side included: a new `CuSimAppLifecycle` mirrors the wrapper with each transition threading the `sim_callback`, and both `cu_rp_balancebot`'s Bevy sim driver and `cu_caterpillar`'s determinism harness now use it. True replay engines (keyframe locking on a started app) deliberately stay on the raw traits behind a short explanatory comment — a resim-aware typestate could be a follow-up if wanted.

Everything in-tree migrated in the same push (examples, component testers, core tests, templates), which kept surfacing real bugs: on top of the six examples that double-started/stopped around `run()`, five hardware testers turned out to be iterating without ever starting their tasks. Eleven latent lifecycle bugs total that this design catches at compile time.

One design point worth your eye: `inner_mut()` is exposed **only** in the `Initialized` state (pre-start configuration, e.g. grabbing the monitor model in the bevymon demo); started states are read-only so the transitions can't be bypassed mid-flight.

Since deprecating `build()` forces the whole-tree migration into the same PR for CI, the examples port is folded in here and #1275 is closed as superseded.

gbin:
[not AI] 
APIs are tough to get right, no worries. This is a better approach. I would love to keep build()... what about exposing both APIs ie. type stateful + the old one on all the states making it deprecated on all the states? Like that this is a no code change for the users + a clear deprecated API warning they can minimally fix?

[heavily prompted, super micromanaged AI]
## Summary

  Rework PR #1272 (https://github.com/copper-project/copper-rs/pull/1272) so the generated builder has one terminal: build(). It returns the boxed lifecycle façade in
  Initialized, while legacy lifecycle calls remain available with consistent deprecation warnings.

  No runtime state enum or hot-path validation is added. Deprecated calls remain an explicit escape from typestate guarantees.

  ## API Changes

  - Replace build_app() with build() -> CuResult<CuAppLifecycle<..., Initialized>>; retain private build_impl() for generated/framework construction.
  - Rename typestate transitions:
      - start(self) -> Running
      - run_one_iteration(&mut self) on Running
      - stop(self) -> Stopped
      - run_until_shutdown(self) -> Stopped
      - Apply equivalent names and callback arguments to CuSimAppLifecycle.

  - Deprecate start_all_tasks, run_one_iteration, run, and stop_all_tasks on CuApplication and CuSimApplication, using matching migration notes and since = "1.2.0".
  - Implement those raw traits for CuAppLifecycle<..., Initialized> and its simulation counterpart, forwarding to the wrapped application. This preserves ordinary, generic, and
    UFCS legacy usage while making every raw lifecycle call warn.

  - Implement Deref in every state and DerefMut only in Initialized, preserving calls such as app.clock() without permitting mutable raw access after a typed start.
  - Keep inner, inner_mut, and into_inner; replay/debug engines requiring unrestricted mid-flight access explicitly unwrap the application.
  - Retain the PR’s boxed storage, faulted-state recovery, restart-from-Stopped, and error conversion behavior.

  Migration becomes:

  let app = MyApp::builder().build()?;
  let mut app = app.start()?;
  app.run_one_iteration()?;
  let app = app.stop()?;

  Legacy code keeps its existing shape and receives warnings. Mixing deprecated and typed methods remains possible and is documented as outside typestate guarantees.

[not AI] Then we will remove those methods with copper 2.0 in a few months

arunvenkatadri:
I'm glad we're converged on approach. I love the idea of not having two method names but didn't know if that was consistent with the Cu philsophy...glad we're on the same page now. 

let' me get with my staff ( agents) and cook up a solution.

gbin:
Ok I think we got it! ... and the deprecation works well!

error: use of deprecated method `cu29::prelude::CuApplication::run_one_iteration`: use `run_one_iteration()` on the Running handle from `build()?.start()?`
   --> core/cu29_runtime/tests/async_keyframes.rs:165:13
    |
165 |         app.run_one_iteration()?;
    |             ^^^^^^^^^^^^^^^^^

maybe we need to disable the warning just for those tests calls so clippy doesn't trip on us for CI/CD.

arunvenkatadri:
Thanks! That was master moving under the PR: the new keyframe tests came in via the merge. Merged master and ported them in a7381a8: plain lifecycle sequences now use the typed handle, and the keyframe-restore tests (which mutate a started app) carry a targeted allow(deprecated) with the same note as replay_primitives.rs, since that's the replay primitive the typestate intentionally doesn't expose. CI should be green now.

PR #1275, opened 2026-08-16
feat(examples): port examples to the CuAppLifecycle typestate API

arunvenkatadri:
Follow-up to #1272, as requested there: ports the examples to the new `CuAppLifecycle` typestate API so the user-side ergonomics are visible.

> **Stacked on #1272.** The base branch lives on my fork, so this PR targets `master` and includes the parent commit — only the last commit (`feat(examples): port examples to the CuAppLifecycle typestate API`, 42 files) is new here. Happy to rebase once #1272 lands.

## What the port looks like

The one-shot case shrinks to a single expression:

```rust
CuStdAppLifecycle::new(application).run()?;
```

The iterate-manually case reads as a state progression, and `?` still works because `LifecycleError` converts into `CuError`:

```rust
let mut running = CuStdAppLifecycle::new(app).start_all_tasks()?;
for _ in 0..ITERATIONS {
    running.run_one_iteration()?;
}
running.stop_all_tasks()?;
```

## Findings from dogfooding/Running the demos. 

**The typestate caught real bugs already in the tree.** Six examples (`cu_monitoring`, `cu_background_task`, `cu_nologging_task`, `cu_rate_target`, `cu_bridge_test`, `cu_resources_test`) called `start_all_tasks()` / `stop_all_tasks()` *around* `run()` — which starts and stops tasks itself, so tasks were double-started and double-stopped at runtime. Under the wrapper that sequence no longer compiles; those examples now use a single `run()` transition.

**UFCS noise disappears.** Call sites that needed `<App as CuApplication<S, L>>::start_all_tasks(&mut app)` to disambiguate (cu_bevymon_demo, cu_bridge_test tests, cu_min_baremetal no_std) become plain method calls because the wrapper pins `S`/`L` once.

**It works in an ECS resource** (`cu_bevymon_demo`): consuming transitions map onto Bevy's `&mut` resources via a small state enum + `mem::replace` swap. The old `started: bool` flag disappears — the state enum *is* the flag. As a bonus the error path now shuts tasks down cleanly instead of leaking them (the old code just flipped `started = false`).

**no_std works** (`cu_min_baremetal`): `CuAppLifecycle::<NoopSectionStorage, NoopLogger, _>::new(app).run()` — verified on `thumbv7em-none-eabihf`.

**Friction worth discussing:**
- The wrapper exposes only `inner()` (shared). That's deliberate — `inner_mut()` would let callers bypass the typestate by calling `start_all_tasks` on the raw app — but it means pre-start mutable setup (e.g. `copper_runtime_mut().monitor.model()` in cu_bevymon_demo) must happen before wrapping. Maybe worth a doc note, or a constrained accessor later.
- On failure paths that only log, the payload is reached through `err.error` (e.g. `debug!("Ended: {}", error.error)`), which reads a little indirect. A `.into_cu_error()` or `Deref` could smooth this if it bothers people.

## Not ported (out of the wrapper's scope)  - Let me know if you want any of this pulled in ( Not AI generated) 

- **sim/resim examples** (`cu_caterpillar` resim/determinism, `cu_rp_balancebot` sim/resim, `cu_run_in_sim`, `cu_debug_session`, `cu_remote_debug_session`, `cu_reflect_demo`, `cu_peer_triangulation`, `cu_flight_controller` sim/resim, `cu_bridge_test/resim`): `CuSimApplication` threads a `sim_callback` through every lifecycle call, which the wrapper doesn't model. A sim-side equivalent could be a follow-up if this API lands.
- **`cu_runtime_matrix`**: its `MissionApp` is a dispatch enum with *inherent* lifecycle methods, not a `CuApplication` implementor.

## Verification

- `cargo check --all-targets` on every ported crate (plus `--features safety-ids` for cu_baremetal_safety and `--no-default-features --target thumbv7em-none-eabihf` for cu_min_baremetal), clippy on a sample, `cargo fmt --check` clean.
- `cargo test --lib`: cu-bridge-test 7/7, cu-baremetal-safety 5/5.
- Locally unverifiable (missing system libs on this machine, unrelated to the change): cu-human-pose (glib), cu-image-codec-demo `ffv1` paths (ffmpeg), cu-python-task-demo link step (python3.9). CI covers these.

### DIFF
--- a/examples/cu_anytime_task/src/main.rs
+++ b/examples/cu_anytime_task/src/main.rs
@@ -201,18 +201,18 @@ fn main() {
     {
         fs::create_dir_all(parent).expect("Failed to create logs directory");
     }
-    let mut application = App::builder()
+    let application = App::builder()
         .with_log_path(logger_path, SLAB_SIZE)
         .expect("Failed to setup logger.")
         .build()
         .expect("Failed to create application.");
-    application
+    let mut running = CuStdAppLifecycle::new(application)
         .start_all_tasks()
         .expect("Failed to start application.");
     // Three copperlists for the foreground chain; keep polling until the
     // backgrounded node's first job comes back from its worker thread.
     for iteration in 0..100 {
-        application
+        running
             .run_one_iteration()
             .expect("Failed to run application.");
         let landed = !tasks::BACKGROUND_RECORDED
@@ -223,7 +223,7 @@ fn main() {
             break;
         }
     }
-    application
+    running
         .stop_all_tasks()
         .expect("Failed to stop application.");
 
--- a/examples/cu_background_task/src/main.rs
+++ b/examples/cu_background_task/src/main.rs
@@ -89,19 +89,18 @@ fn main() {
     }
     debug!("Logger created at {}.", path = &logger_path);
     debug!("Creating application... ");
-    let mut application = App::builder()
+    let application = App::builder()
         .with_log_path(logger_path, SLAB_SIZE)
         .expect("Failed to setup logger.")
         .build()
         .expect("Failed to create application.");
     debug!("Running... starting clock: {}.", application.clock().now());
-    application
-        .start_all_tasks()
-        .expect("Failed to start application.");
-    application.run().expect("Failed to run application.");
-    application
-        .stop_all_tasks()
-        .expect("Failed to stop application.");
-    debug!("End of program: {}.", application.clock().now());
+    // `run` drives the full start/iterate/stop cycle; the typestate wrapper
+    // makes the previous extra start_all_tasks/stop_all_tasks calls around it
+    // a compile error instead of a double start/stop at runtime.
+    let stopped = CuStdAppLifecycle::new(application)
+        .run()
+        .expect("Failed to run application.");
+    debug!("End of program: {}.", stopped.inner().clock().now());
     // check if the logger file is at least 1 section in length
 }
--- a/api/v1/cu29-runtime.txt
+++ b/api/v1/cu29-runtime.txt
@@ -1,5 +1,39 @@
 pub mod cu29_runtime
 pub mod cu29_runtime::app
+pub struct cu29_runtime::app::CuAppLifecycle<S, L, A, State>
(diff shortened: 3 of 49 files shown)


examples/cu_anytime_task/src/main.rs
```diff
@@ -201,18 +201,18 @@ fn main() {
     {
         fs::create_dir_all(parent).expect("Failed to create logs directory");
     }
-    let mut application = App::builder()
+    let application = App::builder()
```
gbin, line 204: Ok I am glad we did this exercise.

This cannot be a wrapper that *adds* to the pile of stuff the user has to worry about, it should be the opposite.

I would have something like an alternative to builder() that would be our future new builder while the old one is deprecated.

examples/cu_background_task/src/main.rs
```diff
@@ -89,19 +89,18 @@ fn main() {
...
-        .stop_all_tasks()
-        .expect("Failed to stop application.");
-    debug!("End of program: {}.", application.clock().now());
+    // `run` drives the full start/iterate/stop cycle; the typestate wrapper
+    // makes the previous extra start_all_tasks/stop_all_tasks calls around it
+    // a compile error instead of a double start/stop at runtime.
+    let stopped = CuStdAppLifecycle::new(application)
+        .run()
+        .expect("Failed to run application.");
+    debug!("End of program: {}.", stopped.inner().clock().now());
```
gbin, line 104: This doesn't look good neither from a user friendliness standpoint.

gbin (changes requested):

gbin:
The main issue is the insertion point of the API as we need to have it as the new default to simplify the user side (be careful with LLM BTW they do have the tendency to bloat stuff like this).

So the fix would be really rethink the app construction flow to use the wrapper instead of just a wrapper on its own the user needs to think about using (more stuff to do for the user instead of less).

My second objection is that in this repo, 2 examples are actual real physical robots and we decided to not port those ones. If there were only 2 examples to port, those would have been the ones.

arunvenkatadri:
Superseded: folded into #1272. Deprecating `build()` there forces the whole-tree migration (examples included) into the same PR for CI, and the examples now use the new `build_app()` default path directly, which removes the wrapper ceremony this PR showcased.

PR #1276, opened 2026-08-17
AnyNet anytime stereo-depth task and demo

Scofield626:
## Summary

Ports the first three stages of [AnyNet](https://github.com/mileyan/AnyNet) (Wang et al., ICRA 2019) as a `CuAnytimeTask`, plus a Rerun demo that makes the latency-versus-quality trade-off visible.

Stage 1 is the base result; stages 2 and 3 are bounded refine quanta. This gives the anytime machinery a real perception workload — the budget genuinely decides how much depth quality you get.

## Related issues
- Closes #

## Changes

**`components/tasks/cu_anynet`** — the task itself, on Candle (optional `cuda` feature).
- Correctness is pinned to the reference implementation: `three_stages_match_the_pytorch_golden_fixture` asserts max error < 1e-4 against a checked-in fixture generated from the authors' PyTorch repo, covering all three stages plus four intermediates. The fixture is committed, so neither Python nor Torch is needed to build or test.
- `stage_snapshots` (default off) keeps a detached copy of what each stage published. Refinement rewrites the depth map in place, so without it only the stage the budget reached survives to the sink. It costs one depth copy per quantum, hence opt-in.
- `normalization` selects input scaling (`imagenet` default, or `raw`). Publicly mirrored checkpoints were trained on raw 0..255 pixels rather than the original `/255` ImageNet protocol, and a mismatch is silent — every stage just regresses confidently wrong disparities.

**`examples/cu_anytime_anynet`** — foreground and background demos.
- The viewer lays the stages out like the paper's architecture figure: disparity panes on a pinned colormap, per-stage residual maps, a pinhole-backed 3D view, and a `time` timeline that replays refinement *within* each frame.
- Given KITTI ground truth it computes the benchmark's 3-pixel error per stage, turning the quality curve from a policy label into a measurement.
- Ships running the synthetic scene with smoke-mode weights, so it works on a fresh clone with no downloads. The README covers opting into KITTI 2015 and a trained checkpoint.
- Both binaries drop their `start_all_tasks()` call — `run()` already starts the tasks, so they were being started twice.

## Reminder

## Additional context

Verified locally on the rebased branch: `just fmt-check`, `just api-check` (baselines unchanged), `cargo clippy --all-targets` clean on both new crates, and `cargo test -p cu-anynet` 11/11. The demo was also run end-to-end against the shipped config with no local data — clean run, valid unified log and `.rrd`.

The reminder box is left unticked deliberately: I ran the format, lint, and API gates plus tests scoped to the new crates, but not the full workspace `just test`. Leaving that to CI.

Two notes for reviewers:
- The README's measured numbers — the ~30% → ~26% → ~15% 3-px error progression and the `time_budget_ms` → published-stage table — are from one desktop CPU and are labelled as such.
- `tools/*.py` are one-shot developer tools, excluded from the published crate. `gen_golden.py` has to be Python: it imports the authors' research repo to produce the conformance oracle, so reimplementing it in Rust would make the test circular. `convert_checkpoint.py` (checkpoint → safetensors) could plausibly move to Rust via `candle_core::pickle`; happy to follow up if preferred.


gbin:
Superseded by #1281, which preserves this commit and adds the maintainer follow-up commit on an in-repository branch.

PR #1277, opened 2026-08-17
feat: pluggable execution-plan scheduling (v1)

Scofield626:
Planners are now injected, monitor-style, instead of picked from a fixed enum.

- Public `CuPlanner` trait: a planner decides only a `StepOrder` over the plan graph; a shared gate rejects illegal orders and materializes the plan.
- RON: `runtime: (planner: (type: "...", config: {...}))`. Copper ships `cu29::planner::Linearity` (default, bit-identical to the historical order — differential test) and `cu29::planner::Pinned` (explicit task order).
- Any crate can provide a planner. Planning is consumed by the macro at compile time, so out-of-tree planners run in the app's `build.rs` via `cu29::planner::emit_plan::<P>(...)`; the macro reads the emitted order and bakes it into the effective config, keeping logs, logstats, and resim self-describing without linking the planner crate.
- `examples/cu_custom_planner` injects an out-of-tree planner end to end and asserts the executed order.

Profile-guided planners (latency objectives) come as follow-up crates on top of this trait.


gbin:
Sorry it took me a while to check that out.

And no this is not the direction I am seeing.

On an API when you want extensibility you should not just give a fixed choice.

Here what we want is the *user* to be able to inject any planner. 
Then we can have a set of planners that are included with copper too.

Look at how the monitors work: you can add any monitors and we give you some, even in different crates if you want to.

We should have the same level of "pluggability" here.

PR #1286, opened 2026-08-24
Quick user friendly bench running + patchs for apple to apple

gbin:
## Related issues
- Closes #

## Reminder


gbin:
@Scofield626 I think we should merge this stack, should I clean it up or there is more you want to test/add to it?

Scofield626:
I am indeed playing with this bench (making sure we are indeed apple-2-apple), and also tried to run natively on Orin. 

The only critical issue is that LAME need to use `SCHED_DEADLINE`. I don't think the prepare script made it correctly.

LaME discards the return value at all four `sched_setattr` sites and keeps printing RT Chain … Response Time after degrading to CFS. In the PR, the calls will fail (root-domain EPERM; 99.898 %/core → EBUSY), and `prepare_lame.py` additionally deletes the affinity block and stubs the cgroup path. Fix is to make every scheduler syscall fatal plus a cpuset partition root?

Another nit thing just to double check is that have you pushed the changes to use multiple threads for copper given LAME uses 4 threads? I think you did, right?

Scofield626 (approved):

gbin:
Yep I also spread copper on 4 cores but we need to fix the scheduling policy otherwise it is not really fair indeed

PR #1290, opened 2026-08-24
feat(cu_flight_controller): MCAP export for both logreaders and scripted sim runs

arunvenkatadri:
## Summary

`cu_flight_controller` could not export its logs to MCAP, and the simulator needed a keyboard to fly. This PR makes the example usable end to end for log analysis: run the sim scripted, export both subsystem logs to MCAP, query them with any MCAP tool.

Motivation: answering "how close did the drone get to a tree?" from the forest-world sim log with an external analysis tool. After this change the answer is a query away (closest ZED depth return over the flight, joined to `/navigation`).

## Related issues
- None filed; happy to open one for the schema tracing note below.

## Changes

- **MCAP export for both logreaders.** `logreader` feature now enables `cu29-export/mcap`; `quad-logreader` and `quad-compute-logreader` implement `PayloadSchemas`. Task payloads with a known type get a traced schema (`NavigationState`, `AutoMissionTarget`, `AutonomyContext`, `AhrsPose`, `ImuPayload`, `ControlInputs`, …); everything else gets an explicit empty-object schema so the channel still exports with Copper's `tov`/`process_time` envelope instead of the export panicking on an unknown id.
- **Explicit depth-map schema.** `trace_type_to_jsonschema::<ZedDepthMap>()` stops at `format` (the `handle` samples and `encoding` are not reflected), so `vitfly_depth_rate` gets a hand-written schema matching the serialized payload: `format`, `encoding{units_per_meter, invalid}`, `handle: integer[]`.
- **Scripted sim runs.** Two optional env vars, no behavior change when unset:
  - `CU_SIM_AUTOSTART_S=<secs>` — arm in Angle mode and engage AUTO once the sim has run that long (equivalent to Space then `4`).
  - `CU_SIM_EXIT_AFTER_S=<secs>` — request a clean `AppExit` after that much physics time, so Copper stops and the unified log is finalized (SIGINT/SIGKILL leave the slab unfinished).

Example:

```bash
CU_SIM_AUTOSTART_S=4 CU_SIM_EXIT_AFTER_S=45 just sim
cargo run -p cu-flight-controller --no-default-features --features logreader --bin quad-logreader -- \
    examples/cu_flight_controller/logs/flight_controller_sim.copper export-mcap --output fc.mcap
cargo run -p cu-flight-controller --no-default-features --features compute-logreader --bin quad-compute-logreader -- \
    examples/cu_flight_controller/logs/flight_compute_sim.copper export-mcap --output compute.mcap
```

On a 45 s forest run: MCU log → 2,880 CopperLists / 115,200 messages / 40 channels in 0.4 s; compute log → 46,080 messages / 16 channels (the `zed` stereo frames dominate the size).

## Reminder
- [x] `cargo fmt`, clippy (`-D warnings`) for the `sim,forest-world` and `compute-logreader` targets, and the `config_matrix` test
- [x] Example updated; no docs/book changes needed

## Additional context

Two things noticed while doing this, left for maintainers to decide:

1. **Unit newtypes trace as objects but serialize as scalars.** `trace_type_to_jsonschema` emits `{"value": number}` for `Length`, `Velocity`, angle types, etc., while serde writes the bare number. Consumers that trust the schema reject every typed channel (`"Could not convert -0.0 with type float ... to struct"`). Downstream I made the consumer tolerant (wrap scalars into single-field structs), but the exporter is the right place to fix it — either trace serde-transparent newtypes as their inner type, or serialize them as objects.
2. `CuDepthMap`'s reflected schema omits `handle`/`encoding`, hence the hand-written schema here; a `Reflect`-visible representation would let the generic trace work.

### DIFF
--- a/examples/cu_flight_controller/Cargo.toml
+++ b/examples/cu_flight_controller/Cargo.toml
@@ -253,6 +253,7 @@ logreader = [
   "bincode/std",
   "serde/std",
   "dep:cu29-export",
+  "cu29-export/mcap",
   "dep:cu-crsf",
   "dep:cu-msp-bridge",
   "dep:cu-bdshot",
--- a/examples/cu_flight_controller/src/compute_logreader.rs
+++ b/examples/cu_flight_controller/src/compute_logreader.rs
@@ -3,10 +3,47 @@ extern crate cu29 as bevy;
 mod messages;
 
 use cu29::prelude::*;
-use cu29_export::run_cli;
+use cu29_export::{PayloadSchemas, run_cli, trace_type_to_jsonschema};
+use messages::*;
 
 // The simulator and deployed compute runtime use the same auto-flying graph.
 gen_cumsgs!("compute_config.ron");
+
+/// Schema for tasks whose payload type is not traced here (see logreader.rs).
+/// `cu_zed::ZedDepthMap` as serialized: 16-bit samples in `handle`, scaled by
+/// `encoding.units_per_meter`, with `encoding.invalid` marking no-return pixels.
+fn depth_map_schema() -> String {
+    r#"{"$schema":"https://json-schema.org/draft-07/schema#","type":"object","properties":{
+"format":{"type":"object","properties":{"width":{"type":"integer"},"height":{"type":"integer"},"stride":{"type":"integer"}}},
+"encoding":{"type":"object","properties":{"units_per_meter":{"type":"number"},"invalid":{"type":"object","properties":{"unsigned":{"type":"integer"}}}}},
+"handle":{"type":"array","items":{"type":"integer"}}}}"#
+        .to_string()
+}
+
+fn untyped_schema() -> String {
+    r#"{"$schema":"https://json-schema.org/draft-07/schema#","type":"object","properties":{}}"#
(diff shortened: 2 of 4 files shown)


gbin:
Hey Arun thanks for the PR. A couple thing: we try to not use environment variables in copper, they are basically hidden parameters and very hard to track. 
Second thing, please try to not bundle things in PRs, they should change one thing. It pays off for rollbacks, let's say we discovered a bug in the mcap export, if we roll back the PR we lose the auto-autonomy feature that was perfectly viable on its own :).

I'll take it from there (split in 2 + move the env variable to the command line).

gbin:
Thank you for putting both improvements together. I split this PR into two focused PRs while preserving the original implementation credit and commit authorship:

- #1291 — compile-time autonomous demo profile for native and browser builds, with no environment, CLI, or URL parameters
- #1292 — MCAP export fix for both flight-controller logreaders

It is better not to accumulate independent changes in a single PR: focused PRs are easier to review and validate, and each change can be merged, reverted, or rolled back independently without affecting the other. This PR is therefore superseded by #1291 and #1292.

PR #1292, opened 2026-08-25
fix(schema): refactor json schema injection on top of reflect. This allows to move cleanly custom schema near type declarations.

gbin:
## Summary

- Enables `cu29-export/mcap` for the flight-controller logreader feature set.
- Makes generated Copper logreaders derive MCAP schemas from the exact payload type of every output slot.
- Removes application-maintained `PayloadSchemas` tables from the flight controller and the other repository logreaders.
- Defines the exceptional depth-map wire schema beside `CuDepthMap` and its custom `Serialize` implementation.

The original implementation and commit authorship remain credited to @arunvenkatadri.

## Design

`gen_cumsgs!` now emits slot-aligned `TaskOutputSpec` metadata that can register each concrete payload type with reflection. The offline MCAP exporter consumes those generated specs directly, so application authors no longer match runtime task-ID strings to payload schemas. Slot ordering is preserved even when one task has several output types sharing the same task ID.

Ordinary payloads use their reflected structure automatically. A reusable payload with a custom Serde wire representation implements the format-neutral `SerializedPayloadSchema` contract beside that serializer and exposes it as reflected type metadata. `CuDepthMap` does this once for its length and compact integer encodings; every current and future logreader then discovers the correct `format`, `encoding`, and `handle` schema automatically.

This keeps ownership in the right places: payload crates describe exceptional wire formats, generated runtime metadata describes statically known output slots, and `cu29_export` owns MCAP envelopes, channels, and timestamps. There is no user-maintained task-name dispatch and no dynamic runtime discovery.

`SerializedPayloadSchema` returns `&'static str`, so the payload-owned schema does not allocate. Reflection registries and owned schema strings are created only in the offline `export-mcap` path. The real-time and embedded paths are unchanged; without `reflect`, output metadata remains static type-name metadata and has no registry function.

The legacy `PayloadSchemas` trait remains exported for source compatibility, but generated `run_cli`/`export_to_mcap` paths no longer require or consult it. Explicit callers can still use the lower-level `export_to_mcap_with_schemas` API.

## Correctness coverage

- Custom serialized schemas override reflected fields, tested with compact `CuDepthMap<Vec<u16>, CuDepthInteger<...>>`.
- Multiple output slots with the same task ID retain different, correctly aligned schemas.
- Exported JSON messages validate against their registered channel schemas.
- All existing repository logreaders compile without manual schema implementations.

## Documentation

- copper-rs-book#41 updates the MCAP guide and removes the manual schema-table instructions.
- The wiki update is pushed as `docs/automatic-mcap-schemas` (`e1230fa`); GitHub does not expose the `.wiki` repository through Pull Requests.

## Verification

- `just fmt`
- `just lint` (host workspace, core `no_std`, and embedded target clippy checks)
- `just api-check`
- `cargo test -p cu29-export --features mcap` (26 passed)
- `cargo test -p cu-sensor-payloads` (37 unit tests plus integration/doc tests passed)
- `cargo test -p cu29-runtime --features reflect -- --test-threads=1`
- `cargo check -p cu-dorabench -p cu-caterpillar -p cu-min-baremetal -p cu-parallel-mandelbrot -p cu-nologging-task -p cu-rp-balancebot --bins`
- `cargo clippy -p cu-flight-controller --no-default-features --features compute-logreader --bin quad-logreader --bin quad-compute-logreader -- -D warnings`


gbin:
@arunvenkatadri I think that is a better generic direction to inject any custom json schema, tell me what you think.

core/cu29_export/src/mcap_export.rs
```diff
@@ -351,8 +361,10 @@ where
         )
         .map_err(|e| CuError::new_with_cause("Failed to add metadata schema", e))?;
 
-    for task_id in task_ids.iter() {
-        if let Some(schema_json) = schema_map.get(task_id) {
+    for (slot_index, task_id) in task_ids.iter().enumerate() {
```
Scofield626, line 374: this might be problematic:

Master built schema_map: BTreeMap<&str,&str>; the PR uses schemas.get(slot_index) plus a name equality check. Unordered or deduplicated input now silently falls to add_channel(0, ...). The doc still says "pairs".

components/payloads/cu_sensor_payloads/src/depth.rs
```diff
@@ -505,6 +513,41 @@ where
...
+            r#"},"required":["units_per_meter","invalid"],"additionalProperties":false},
+"handle":{"type":"array","items":"#,
+            $sample_schema,
+            r#"}},"required":["format","encoding","handle"],"additionalProperties":false}"#
+        )
+    };
+}
+
+const LENGTH_DEPTH_MAP_SERIALIZED_SCHEMA: &str =
+    depth_map_serialized_schema!(r#"{"type":"number"}"#, r#"{"type":"null"}"#);
```
Scofield626: CuDepthLength::encode_sample(None) stores f32::NAN (depth.rs:95), and serde_json writes non-finite floats as null. Any depth map with one invalid pixel fails its own schema. Use {"type":["number","null"]}.

Scofield626:
Overall LGTM!

core/cu29_export/src/mcap_export.rs
gbin: ok the pairs is more like ordered pairs per slot, let me fix this

PR #1301, opened 2026-09-02
Fix PNG log codec record framing

gbin:
## Summary

- frame each encoded PNG as a bincode length-delimited byte buffer
- reuse codec-owned scratch storage across encodes
- add a multi-frame regression test

## Why

The previous streaming decoder could read past the PNG IEND chunk into the following CopperList record, desynchronizing multi-record logs. Since that format could not reliably decode real logs, this fixes the framing directly without adding a format version.

## Validation

- cargo +stable test -p cu-png-codec
- cargo +stable clippy --release -p cu-png-codec -- -D warnings
- just fmt

### DIFF
--- a/components/codecs/cu_png_codec/src/lib.rs
+++ b/components/codecs/cu_png_codec/src/lib.rs
@@ -1,6 +1,4 @@
-use bincode::de::read::Reader as BincodeReader;
 use bincode::de::{Decode, Decoder};
-use bincode::enc::write::Writer as BincodeWriter;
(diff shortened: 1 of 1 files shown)


gbin:
Superseded by #1380. Copper now reserves a fixed-size payload length, lets the codec encode directly into log storage, and fills in the actual length afterward. Bounded decoding protects adjacent payloads and metadata. This replaces the PNG-specific encoded scratch buffer and per-frame compressed-data allocation proposed here.

PR #1302, opened 2026-09-03
Feat(fsck): Report CopperList byte entropy in fsck

gbin:
## Summary
- compute zero-order Shannon entropy over used CopperList section bytes
- report the idealized entropy floor and remaining byte-frequency headroom
- handle logs with no CopperList bytes

This is super useful to understand if you need codecs for your datatypes.

Example:

    CL byte entropy  -> 7.0975 bits/byte
    CL entropy floor -> 1,874,380 bytes (11.28% zero-order headroom)

## Validation
- just fmt
- cargo test -p cu29-export fsck::tests
- cargo clippy -p cu29-export --all-targets -- -D warnings

### DIFF
--- a/core/cu29_export/src/fsck.rs
+++ b/core/cu29_export/src/fsck.rs
@@ -8,6 +8,51 @@ use cu29::{CopperListTuple, CuResult};
 use num_format::{Locale, ToFormattedString};
 use std::io::Cursor;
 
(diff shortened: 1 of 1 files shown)


gbin:
This metric was motivated and validated by the before/after traces in copper-project/copper-rs#1303.

PR #1313, opened 2026-09-04
fix(spatial): pin row-major layout for Transform3D from_matrix/to_matrix

### LINKED ISSUE #1255: Transform3D: translation()/rotation() return zeros/transposed values under the default glam feature (row-major vs column-major mismatch)
`Transform3D` mixes two matrix layouts, and the unit accessors are broken whenever the `glam` feature is on (the default).

## What happens

The glam backend converts through `from_cols_array_2d`/`to_cols_array_2d`, which treat each inner array as a **column** — so `from_matrix`/`to_matrix` speak column-major, with the translation at `mat[3][0..3]`:

https://github.com/copper-project/copper-rs/blob/72eaef6350ec9e81cda2447b4e54068e15e5da96/components/payloads/cu_spatial_payloads/src/lib.rs#L230

https://github.com/copper-project/copper-rs/blob/72eaef6350ec9e81cda2447b4e54068e15e5da96/components/payloads/cu_spatial_payloads/src/lib.rs#L248-L252

But `translation()` reads the **row-major** cells `mat[i][3]` — always 0 in a column-major affine — so it returns zeros for every transform, and `rotation()` returns the transposed (inverse) rotation:

https://github.com/copper-project/copper-rs/blob/72eaef6350ec9e81cda2447b4e54068e15e5da96/components/payloads/cu_spatial_payloads/src/lib.rs#L85-L98

## Repro

```rust
let t = Transform3D::<f32>::from_matrix([
    [1., 0., 0., 0.],
    [0., 1., 0., 0.],
    [0., 0., 1., 0.],
    [2., 3., 4., 1.],
]);
assert_eq!(t.to_matrix()[3], [2., 3., 4., 1.]); // lossless round-trip
t.translation(); // glam (default): [0, 0, 0] — no-default-features: [2, 3, 4]
```

## Consequences

- `translation()`/`rotation()` are unusable in default builds.
- The non-glam path is consistently row-major (its `inverse()` reads `mat[i][3]`: https://github.com/copper-project/copper-rs/blob/72eaef6350ec9e81cda2447b4e54068e15e5da96/components/payloads/cu_spatial_payloads/src/lib.rs#L385-L388), so the two feature configs serialize **different byte layouts for the same pose** — logs written by a glam build replay wrong in a non-glam build and vice versa.
- The existing tests miss it because both sides of each assertion pass through the same lossy `from_matrix` interpretation, so they degenerate identically under glam (e.g. the inverse tests compare identity against identity once glam drops the row-major translation).

## Suggested fix

Pin one canonical layout for `from_matrix`/`to_matrix`/serialization (row-major matches the non-glam code and the doc examples), transpose at the glam boundary, and read `translation()`/`rotation()` from the affine directly. #1254 already avoids `to_matrix()` for the new `position()`/`transform_point()`/`transform_vector()` for this reason; happy to follow up with the fix PR.

webzuweb:
## Problem

Closes #1255.

`Transform3D::from_matrix`/`to_matrix` speak **column-major** when the `glam` feature is on (the default), while the non-glam backend is **row-major**. That means:

- A row-major matrix fed to `from_matrix` is misread under glam (translation ends up in the wrong place, rotation transposed).
- `to_matrix` returns a different byte layout depending on the backend, so **logs serialized by a glam build replay wrong in a non-glam build** (and vice versa).
- The existing `From<Transform3D<f32>> for Affine3A` read translation from the last column while `From<Transform3D<f64>> for DAffine3` read it from the last row — the two impls disagreed with each other.

## Fix

Pin **row-major** as the single canonical layout (matching the non-glam backend and the doc examples):

- Transpose at the glam boundary in `TransformInner::from_matrix`/`to_matrix`.
- Rewrite the `glam` `From`/`Into` impls to use the same convention.
- Update downstream consumers in `cu-transform` (interpolation, velocity computation, test utilities) and the rerun integration to read translation from the **last column**.
- Add a row-major round-trip regression test (`from_matrix_is_row_major_and_round_trips`).

## Verification

```
cargo test -p cu-spatial-payloads             # 27 passed (glam)
cargo test -p cu-spatial-payloads --no-default-features  # 25 passed (non-glam)
cargo test -p cu-transform                     # all passed
```

Both backends now serialize the same byte layout for the same pose.

### DIFF
--- a/components/libs/cu_transform/examples/basic_usage.rs
+++ b/components/libs/cu_transform/examples/basic_usage.rs
@@ -167,7 +167,7 @@ fn main() {
             let mat = transform.to_matrix();
             println!(
                 "Retrieved transform: translation=({:.2}, {:.2}, {:.2})",
-                mat[3][0], mat[3][1], mat[3][2]
+                mat[0][3], mat[1][3], mat[2][3]
             );
         }
         Err(e) => println!("Error: {e}"),
--- a/components/libs/cu_transform/examples/payload_usage.rs
+++ b/components/libs/cu_transform/examples/payload_usage.rs
@@ -104,7 +104,7 @@ fn main() {
             let mat = transform.to_matrix();
             println!(
                 "  World->arm at t=1s: translation=({:.2}, {:.2}, {:.2})",
-                mat[3][0], mat[3][1], mat[3][2]
+                mat[0][3], mat[1][3], mat[2][3]
             );
         }
         Err(e) => println!("  Error: {e}"),
--- a/components/libs/cu_transform/src/interpolation.rs
+++ b/components/libs/cu_transform/src/interpolation.rs
@@ -121,12 +121,12 @@ pub fn interpolate_transforms<T: Interpolate>(
         }
     }
 
-    // Interpolate translation (in column-major format, translation is in the last row)
+    // Interpolate translation (row-major: translation is in the last column).
     for i in 0..3 {
-        let before_val = before_mat[3][i].to_f64();
-        let after_val = after_mat[3][i].to_f64();
+        let before_val = before_mat[i][3].to_f64();
+        let after_val = after_mat[i][3].to_f64();
         let interpolated = before_val + (after_val - before_val) * ratio;
-        result_mat[3][i] = T::from_f64(interpolated);
+        result_mat[i][3] = T::from_f64(interpolated);
     }
 
     Ok(Transform3D::from_matrix(result_mat))
@@ -166,10 +166,10 @@ mod tests {
 
         let after: StampedTransform<f32> = StampedTransform {
             transform: Transform3D::from_matrix([
-                [1.0, 0.0, 0.0, 0.0],
+                [1.0, 0.0, 0.0, 10.0],
                 [0.0, 1.0, 0.0, 0.0],
                 [0.0, 0.0, 1.0, 0.0],
-                [10.0, 0.0, 0.0, 1.0], // Translation: x=10
+                [0.0, 0.0, 0.0, 1.0], // Translation: x=10
             ]),
             stamp: CuTime::from_nanos(3000),
             parent_frame: frame_id!("world"),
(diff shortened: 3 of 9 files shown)


gbin:
we have a leakage of storage format in the API and it looks like we are bound to flip flop for ever:
- Copper API: m[row][column], column vectors, translation in m[0..3][3].
  - glam/Rerun: column-major storage, so adapters transpose/repack.
  - nalgebra/faer: index by (row, column); storage order is irrelevant when copying element-wise.
  - no-glam: raw nested row arrays.

webzuweb:
Agreed — the format leakage is precisely the bug this PR targets. `from_matrix`/`to_matrix` were exposing the backend's *native* layout instead of a fixed convention, so the same logical pose produced a different byte layout depending on whether `glam` was compiled in. That's what broke #1255: logs serialized by a `glam` build replay wrong in a non-glam build (and vice versa).

The PR pins **row-major** (`m[row][col]`, translation in `m[0..3][3]`) as the single canonical layout — matching the documented Copper API — and confines every backend-specific transpose/repack to the boundary:

- `glam`: `Mat4::from_cols_array_2d(&m).transpose()` at the `from_matrix`/`to_matrix` boundary
- `nalgebra`/`faer`: element-wise copy, layout-agnostic
- `no-glam`: raw nested row arrays, already row-major

Net effect: `from_matrix`/`to_matrix` round-trip identically on both backends (covered by a new regression test), and downstream consumers read translation from the last column everywhere.

Happy to flip it to **column-major** if you'd rather the canonical follow `glam` (the default) — the important part is that there's exactly *one* convention instead of two. Or, if you'd prefer a fully storage-agnostic API (explicit row/col accessors instead of raw `[[T;4];4]`), I can rework it in that direction too.

gbin:
Looks good to merge. Please clarify the public matrix convention in these docs.
`from_matrix`:
```rust
    /// Creates a homogeneous 4x4 transform indexed as `mat[row][column]`.
    /// Each inner array is a row; points are column vectors (`p_out = mat * p_in`).
    /// Translation is `[mat[0][3], mat[1][3], mat[2][3]]`; the last row must be `[0, 0, 0, 1]`.
    /// The convention is backend-independent: callers must not transpose for glam.
    /// Glam/Rerun adapters repack into columns internally; the fallback stores rows directly.
    /// Nalgebra/faer adapters copy by `(row, column)`, independent of storage order.
```

`to_matrix`:
```rust
    /// Returns a homogeneous 4x4 matrix indexed as `mat[row][column]`.
    /// Each inner array is a row, with translation in the last column, regardless of backend.
    /// Uses the same column-vector convention as [`Self::from_matrix`].
```

webzuweb:
Done — added the row-major convention docs to `from_matrix` and `to_matrix` exactly as suggested. Pushed to the branch; regression tests still green (27 passed, glam + no-glam).

gbin:
thanks

PR #1333, opened 2026-09-07
fix(runtime): key planner output ports on (msg_type, src_channel)

### LINKED ISSUE #791: `compute_runtime_plan` conflates channel-distinct outputs when keyed only by `msg_type`
## Summary

Runtime planning identifies source output ports by `msg_type` only.
If a source has multiple outputs with the same `msg_type` but different `src_channel`, those routes collapse to one planner port identity.

## Evidence

In `core/cu29_runtime/src/curuntime.rs`:
- `collect_output_msg_types(...)` deduplicates by `edge.msg`.
- input `src_port` resolution uses `.position(|msg| msg == msg_type)`.

So planner identity is effectively `msg_type`, not `(msg_type, src_channel)`.

## Why `cu_elrs_bdshot_demo` still works

`examples/cu_elrs_bdshot_demo` uses `#[copper_runtime(...)]`, which goes through `cu29_derive`, not raw graph planning.

`cu29_derive` first normalizes bridge channels into synthetic per-channel nodes (`__CuBridgeRxChannel` / `__CuBridgeTxChannel`) and only then calls `compute_runtime_plan` on that normalized graph.
That preserves channel separation for runtime execution.

## Where this still impacts behavior

Callers that run `compute_runtime_plan` directly on the raw config graph can still observe channel conflation, e.g.:
- `core/cu29_export/src/logstats.rs`
- `components/monitors/cu_consolemon/src/lib.rs`

## Potential fix

Make planner routing keys channel-aware:
1. Use `(msg_type, src_channel)` as output route identity.
2. Deduplicate and resolve `src_port` with that key.
3. Keep fanout for identical full keys.

webzuweb:
## Summary

Runtime planning resolved a source's output ports by `msg_type` only. A source exposing several outputs with the same message type on different bridge channels (e.g. a stereo driver publishing `Image` on `left` and `right`) collapsed them onto a single planner port, so downstream inputs were routed to the wrong (first) port.

Fixes #791.

## Changes

- Add `CuGraph::get_node_output_ports_by_id`, a channel-aware variant of `get_node_output_msg_types_by_id` that keeps channel-distinct ports `(msg_type, src_channel)` and preserves their relative ordering.
- Carry a parallel `src_channels: Vec<Option<String>>` in `CuOutputPack`.
- Resolve `src_port` in `collect_step_inputs` by matching both `msg_type` and `src_channel`.

The derive path (`cu29_derive`) is unaffected: it already normalizes bridge channels into synthetic per-channel nodes before planning.

## Testing

- New regression test `test_runtime_plan_distinguishes_channel_distinct_outputs` verifies a source with two same-typed channel outputs produces two distinct ports and the sink resolves `src_port` 0 and 1.
- `cargo test -p cu29-runtime --lib`: 246 passed, 0 failed.
- `cargo check -p cu29-export -p cu29-derive`: clean.

### DIFF
--- a/api/v1/cu29-runtime.txt
+++ b/api/v1/cu29-runtime.txt
@@ -323,6 +323,7 @@ pub fn cu29_runtime::config::CuGraph::get_node_mut(&mut self, node_id: cu29_runt
 pub fn cu29_runtime::config::CuGraph::get_node_output_msg_type(&self, node_id: &str) -> core::option::Option<alloc::string::String>
 pub fn cu29_runtime::config::CuGraph::get_node_output_msg_types(&self, node_id: &str) -> core::option::Option<alloc::vec::Vec<alloc::string::String>>
 pub fn cu29_runtime::config::CuGraph::get_node_output_msg_types_by_id(&self, node_id: cu29_runtime::config::NodeId) -> cu29_traits::CuResult<alloc::vec::Vec<alloc::string::String>>
+pub fn cu29_runtime::config::CuGraph::get_node_output_ports_by_id(&self, node_id: cu29_runtime::config::NodeId) -> cu29_traits::CuResult<alloc::vec::Vec<(alloc::string::String, core::option::Option<alloc::string::String>)>>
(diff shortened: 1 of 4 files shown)


gbin:
thanks

PR #1365, opened 2026-09-08
feat(cu-ros2-payloads): add geometry_msgs, nav_msgs and CompressedImage/CameraInfo

edgarriba:
## Summary

`cu-ros2-payloads` covers `sensor_msgs`, `std_msgs` scalars and `builtin`. A graph that publishes a
transform, an odometry, a path or a compressed camera stream currently has to define those wire
structs downstream and hand-paste their RIHS01 type hashes — which is silent to get wrong, since the
hash goes into the rmw_zenoh key expression and a mismatch just means the subscriber never matches,
with no error on either side.

This adds the missing message types and gives every hash a single home.

## Related issues
- Closes #

## Changes

**New message types**

- `geometry_msgs`: `Point`, `Pose`, `PoseStamped`, `Transform`, `TransformStamped`, `Twist`,
  `PoseWithCovariance`, `TwistWithCovariance`, plus `Vector3` and `Quaternion`.
- `nav_msgs`: `Odometry`, `Path`.
- `sensor_msgs`: `CompressedImage`, `CameraInfo`, `RegionOfInterest`.

**`Vector3` / `Quaternion` move to `geometry_msgs`, re-exported from `sensor_msgs`**

ROS 2 homes both in `geometry_msgs`, and the namespace is not cosmetic — it feeds `namespace()` and
the key expression. They are defined in `geometry_msgs` now and `sensor_msgs` keeps
`pub use crate::geometry_msgs::{Quaternion, Vector3};`, so existing `sensor_msgs::Vector3` paths
still compile and `Imu` and `Twist` share one type rather than two structurally identical ones.

**A `RosMessage` trait for wire identity**

```rust
pub trait RosMessage {
    const NAMESPACE: &'static str;
    const TYPE_NAME: &'static str;
    const TYPE_HASH: &'static str;
}
```

The three things a bridge puts in an rmw_zenoh key expression, kept next to the struct instead of
copied into every adapter that publishes it. The existing `sensor_msgs` adapters now read their
constants from it; **the literals are byte-identical to what they returned before**. `TYPE_HASH` is
the Jazzy RIHS01 hash and needs no distro `cfg`, because the blanket `RosBridgeAdapter` impl already
substitutes `"TypeHashNotSupported"` under `humble`.

Only types whose hash is verified against a ROS 2 Jazzy install implement the trait. Sub-messages
that are never published on their own (`Point`, `Pose`, `Transform`, `RegionOfInterest`, ...)
deliberately do not, rather than carry a guessed literal.

**A `fixed_array` serde helper (no new dependency)**

`PoseWithCovariance` and `TwistWithCovariance` carry `float64[36]`. serde's built-in `[T; N]` impls
stop at `N == 32`, and encoding the matrix as a `Vec` would prepend a `uint32` length that no ROS
subscriber expects, shifting every field after it. ~30 lines of `serialize_tuple` /
`deserialize_tuple` rather than pulling in `serde_arrays`.

**`Default` on `Header` and `Time`**

Needed so the new messages can derive `Default`. Documented as the ROS default (zero stamp, empty
`frame_id`) with a note that both are valid on the wire and neither is useful — a zero stamp turns
pipeline latency into apparent jitter downstream, and an empty frame builds no tf tree.

**Tests**

CDR round-trips for every new message, plus byte-count assertions for the property that is easy to
break and invisible when broken: fixed arrays carry no length prefix. `camera_info_d_is_a_sequence_while_k_r_p_are_fixed`
pins both halves of that distinction. A cross-module test checks every `TYPE_HASH` is a well-formed
lowercase RIHS01 digest and that no two types share one.

## Reminder
- [x] I have updated docs or examples where needed (rustdoc on the new types; no example changes needed)

## Additional context

**On the `just` checkbox** — I could not run the full workspace gate from a clean clone:
`examples/cu_flight_controller` path-depends on `cu-zed` at `../../../zed/components/sources/cu_zed`,
a sibling repo that is not part of this one, so `cargo` fails to load the workspace manifest before
any check runs:

```
error: failed to load manifest for workspace member `examples/cu_flight_controller`
Caused by: failed to read `.../zed/components/sources/cu_zed/Cargo.toml`
```

Excluding that example locally (not part of this diff), I ran the crate-scoped equivalents:

- `cargo test -p cu-ros2-payloads --features jazzy` — 27 passed
- `cargo test -p cu-ros2-payloads --features humble` — 27 passed
- `cargo clippy -p cu-ros2-payloads --features jazzy --all-targets` — clean
- `cargo fmt -p cu-ros2-payloads` — no diff
- `cargo build -p cu-ros2-bridge --features jazzy` — builds against the change

Happy to adjust anything here; flagging the `zed` path dep separately in case a clean-clone
workspace load is meant to work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Gn1oe6gBfhFHenJBbo7nXN

### DIFF
--- a/components/payloads/cu_ros2_payloads/src/builtin.rs
+++ b/components/payloads/cu_ros2_payloads/src/builtin.rs
@@ -1,13 +1,20 @@
 use compact_str::CompactString;
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+/// `std_msgs/Header`.
+///
+/// `Default` is the ROS default: a zero stamp and an empty `frame_id`. Both are valid on the wire
+/// and neither is useful — a zero stamp turns pipeline latency into apparent jitter downstream,
+/// and an empty frame builds no tf tree — so fill them from the payload's own capture time and
+/// frame rather than defaulting.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
 pub struct Header {
     pub stamp: Time,
     pub frame_id: CompactString,
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+/// `builtin_interfaces/Time`.
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, PartialEq)]
 pub struct Time {
     pub sec: i32,
     pub nanosec: u32,
--- a/components/payloads/cu_ros2_payloads/src/fixed_array.rs
+++ b/components/payloads/cu_ros2_payloads/src/fixed_array.rs
@@ -1,0 +1,95 @@
+//! CDR-correct serde for fixed-size arrays longer than serde's built-in limit.
+//!
+//! `serde` implements `Serialize`/`Deserialize` for `[T; N]` only up to `N == 32`, which is one
+//! short of the 6x6 covariance matrices ROS uses (`float64[36]`). Encoding one as a `Vec` is not
+//! an option: in CDR a *fixed* array is written as bare elements, while a *sequence* is prefixed
+//! with a `uint32` length, so a `Vec` would insert four bytes that no ROS subscriber expects and
+//! shift every field after it.
+//!
+//! Use it as `#[serde(with = "crate::fixed_array")]` on any `[T; N]` field.
+
+use core::fmt;
+use core::marker::PhantomData;
+use serde::de::{Deserialize, Deserializer, Error as DeError, SeqAccess, Visitor};
+use serde::ser::{Serialize, SerializeTuple, Serializer};
+
+pub fn serialize<S, T, const N: usize>(array: &[T; N], serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+    T: Serialize,
+{
+    // `serialize_tuple`, not `serialize_seq`: a tuple has a statically known length, which is what
+    // makes the CDR backend omit the length prefix.
+    let mut tuple = serializer.serialize_tuple(N)?;
+    for element in array {
+        tuple.serialize_element(element)?;
+    }
+    tuple.end()
+}
+
+struct ArrayVisitor<T, const N: usize>(PhantomData<T>);
+
+impl<'de, T, const N: usize> Visitor<'de> for ArrayVisitor<T, N>
+where
+    T: Deserialize<'de>,
+{
+    type Value = [T; N];
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "an array of length {N}")
+    }
+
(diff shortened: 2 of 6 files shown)


gbin:
Nice clean up and added support. 

Maybe as a follow up PR:

Our established pattern, in cu_ros2_payloads (components/payloads/cu_ros2_payloads/src/lib.rs:74), is:

  - From<&CopperPayload> for RosType for outbound conversion.
  - TryFrom<RosType> for CopperPayload for inbound conversion.
  - RosMsgAdapter on the Copper payload, specifying the ROS output type and identity.
  - A blanket implementation then supplies RosBridgeAdapter for bidirectional transport.

  Existing examples include CuImage<Vec<u8>> ↔ Image, ImuPayload ↔ Imu, point clouds ↔ PointCloud2, and primitives ↔ std_msgs.

  The new RosMessage trait alone does not provide bridge integration. 
  Can you at least build the conversions where existing types overlap—particularly Point3f/Point3d ↔ Point, and assess Transform3D ↔
  Transform/Pose with explicit handling of non-rigid transforms? 
  Stamped variants also need a deliberate source for timestamps and frame IDs.

edgarriba:
Thanks @gbin — done in #1366, stacked on this one (draft until this merges; only the second commit is new).

It follows the four-part shape you described:

- `Point3f`/`Point3d` ↔ `geometry_msgs/Point`, with the inbound f64→f32 narrowing erroring instead of saturating to infinity.
- `Transform3D<f32>`/`<f64>` ↔ `Transform`. Non-rigid is a hard error, not a projection: `validate_ros_message` checks RᵀR = I and det = +1, and since the bridge calls that hook before conversion, a transform carrying scale or shear fails before anything reaches the wire. Silently dropping a scale would publish a plausible rigid transform that nothing downstream could detect.
- `Pose` gets the conversions but no adapter — `Output` is an associated type and `cu_spatial_payloads` aliases `Pose<T> = Transform3D<T>`, so one type cannot adapt to both spellings. The adapter spends it on `Transform`.

On the stamped variants: I did not bind them, because there is no deliberate source to bind them to. Nothing in these payloads carries a capture time or a frame id, and an invented header is worse than no message — it is the same trap as `default_header()` in `sensor_msgs.rs`, which stamps `{0,0}`/`""` on all four existing outbound conversions. I sketched two options in the #1366 description (a `Stamped<T>` payload, or frame_id from channel config + stamp from the `CuMsg` tov, which needs metadata threaded into `encode_payload_with`) — happy to implement whichever you prefer.

One correction to this PR's description: I claimed the sub-message hashes could not be verified. They can — `all_rihs.md` is right here at the crate root and I missed it. #1366 binds the remaining types from it and adds a test checking every `TYPE_HASH` against that table. All 20 agree, including the four that predate both PRs.

PR #1366, opened 2026-09-08
feat(cu-ros2-payloads): bridge cu_spatial_payloads to geometry_msgs

edgarriba:
## Summary

Follow-up to #1365, answering @gbin's review there: that PR added the wire structs but no bridge
integration, so this wires the Copper spatial payloads that overlap them.

> **Stacked on #1365 — draft until that merges.** Cross-fork PRs cannot take a base branch that
> lives on the fork, so GitHub shows both commits here. Only the second commit
> (`feat(cu-ros2-payloads): bridge cu_spatial_payloads to geometry_msgs`) is new. I will rebase and
> mark this ready as soon as #1365 lands.

## Related issues
- Closes #

## Changes

Follows the four-part shape the crate already establishes: `From<&Payload> for RosType` outbound,
`TryFrom<RosType> for Payload` inbound, `RosMsgAdapter` for identity, blanket impl supplying
`RosBridgeAdapter`.

**`Point3f` / `Point3d` ↔ `geometry_msgs/Point`**

Widening f32 → f64 is exact. Inbound, f64 → f32 **errors** instead of saturating: `1e300 as f32` is
`inf`, so the silent path would place the point at infinity rather than report that it does not fit.

**`Transform3D<f32>` / `Transform3D<f64>` ↔ `geometry_msgs/Transform`**

Rotation extracted by Shepperd's method (branch on the largest divisor) so the square root never
approaches zero, which is where the trace-only formula loses its precision.

**Non-rigid transforms are refused, not flattened** — the explicit handling you asked for.

`Transform3D` is a full 4×4 affine and can carry scale, shear or a reflection. `geometry_msgs/Transform`
is a translation plus a unit quaternion and can carry none of them. Converting anyway publishes a
plausible-looking rigid transform with the scale quietly dropped, and nothing downstream can detect
that. So `validate_ros_message` checks `RᵀR = I` (no scale, no shear) and `det = +1` (no reflection),
and the bridge already calls that hook before conversion — `cu_ros2_bridge/src/lib.rs:88` — so it
fails before anything reaches the wire. Tolerances are `1e-9` (f64) / `1e-5` (f32): above the drift
of a long rotation chain, below any real scale factor.

Inbound, a quaternion off the unit sphere is rejected rather than normalized — normalizing hides the
sender's bug.

**`Pose` — assessed, and the type system decides it**

`Pose` gets the same conversions but no adapter. `Output` is an associated type, so a type can carry
only one `RosMsgAdapter`, and `cu_spatial_payloads` aliases `Pose<T> = Transform3D<T>` — the *same*
type, not a distinct one. The adapter spends its single `Output` on `Transform`, the name ROS gives
the same concept; `From<&Transform3D<T>> for Pose` and `TryFrom<Pose>` stay available for a graph
that wants the pose spelling.

**Stamped variants: deliberately still unbound**

You asked for a deliberate source for timestamps and frame IDs. There isn't one yet, so I have not
bound `PoseStamped` / `TransformStamped`. No payload here carries a capture time or a frame id, so an
adapter could only invent both — and an invented header is worse than no message: a zero stamp turns
pipeline latency into apparent jitter for every consumer, and an empty `frame_id` silently builds no
tf tree. (This is the same trap as `sensor_msgs.rs`'s `default_header()`, which stamps `{0,0}` and
`""` on all four existing outbound conversions.)

Two ways I can see to give it a real source, happy to implement whichever you prefer:

1. **A `Stamped<T>` payload** in this crate carrying `CuTime` + `frame_id` alongside the value, so the
   header is explicit at the graph level and the conversion has something honest to read.
2. **Bridge-level**: `frame_id` from the channel config, stamp from the `CuMsg` `tov` — which means
   threading metadata into `encode_payload_with`, currently payload-only.

**Also: the type hashes are now verified rather than trusted**

In #1365 I left sub-messages without `RosMessage` rather than paste a guessed hash. That was too
cautious — this repo already ships the answer in `all_rihs.md` (489 rows from the Jazzy IDL), which I
had missed at the crate root. `Point`, `Pose`, `Vector3`, `Quaternion`, `Transform`,
`PoseWithCovariance`, `TwistWithCovariance` and `RegionOfInterest` now carry hashes taken from it,
and a new test checks **every** `TYPE_HASH` in the crate against that table via `include_str!`. All
20 agree today, including the four that predate both PRs. The literals are no longer unverifiable.

No change to `cu_ros2_bridge`: it registers only primitives by default and exposes
`register_ros2_payload::<T>()` for the rest, which is how these become usable.

## Reminder

## Additional context

Same `just` caveat as #1365 — `examples/cu_flight_controller` path-depends on a sibling `zed` repo,
so a clean clone cannot load the workspace manifest. Excluding it locally (not part of this diff):

- `cargo test -p cu-ros2-payloads --features jazzy` — 36 passed
- `cargo test -p cu-ros2-payloads --features humble` — 36 passed
- `cargo clippy -p cu-ros2-payloads --features jazzy --all-targets` — clean
- `cargo fmt -p cu-ros2-payloads` — no diff
- `cargo build -p cu-ros2-bridge --features jazzy` — builds

New tests cover the round trips, the f32 overflow and NaN rejections, a scaled transform refused, a
reflection refused, a non-unit quaternion refused, and that the `Pose` and `Transform` spellings
agree.

The dependency is `default-features = false`, matching `cu_gnss_payloads` and `cu_rrt_star`, so ROS
users do not inherit glam — the `Transform3D` accessors used here are backend-independent.

### DIFF
--- a/components/payloads/cu_ros2_payloads/src/geometry_msgs.rs
+++ b/components/payloads/cu_ros2_payloads/src/geometry_msgs.rs
@@ -209,13 +209,42 @@ use cu_spatial_payloads::{Point3d, Point3f, Transform3D};
 use cu29::units::si::f32::Length as Length32;
 use cu29::units::si::f64::Length as Length64;
 use cu29::units::si::length::meter;
+use glam::{DAffine3, DMat3, DQuat, DVec3};
 
-/// How far a rotation block may stray from orthonormality before it is rejected.
+/// How far a rotation may stray from orthonormality before it is rejected.
 ///
 /// A rotation composed repeatedly drifts, so the check cannot demand exactness; these are set
 /// well above the accumulated error of a long chain and well below any real scale factor.
 const RIGID_TOLERANCE_F64: f64 = 1e-9;
 const RIGID_TOLERANCE_F32: f64 = 1e-5;
+
+/// Read a `Transform3D`'s rotation block as a glam matrix.
+///
+/// `Transform3D::rotation` returns rows (`r[row][column]`) and glam is column-major, so this is
+/// where the two conventions meet. Everything downstream is glam's.
+fn rotation_matrix<T>(rotation: [[T; 3]; 3]) -> DMat3
+where
+    f64: From<T>,
+    T: Copy,
+{
+    DMat3::from_cols(
+        DVec3::new(
+            f64::from(rotation[0][0]),
+            f64::from(rotation[1][0]),
+            f64::from(rotation[2][0]),
+        ),
+        DVec3::new(
+            f64::from(rotation[0][1]),
+            f64::from(rotation[1][1]),
+            f64::from(rotation[2][1]),
+        ),
+        DVec3::new(
+            f64::from(rotation[0][2]),
+            f64::from(rotation[1][2]),
+            f64::from(rotation[2][2]),
+        ),
+    )
+}
 
 /// Reject a rotation block that is not a pure rotation.
 ///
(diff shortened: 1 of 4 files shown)


components/payloads/cu_ros2_payloads/src/geometry_msgs.rs
```diff
@@ -0,0 +1,880 @@
...
+             of -1 is a reflection, which geometry_msgs/Transform cannot represent."
+        ));
+    }
+
+    Ok(())
+}
+
+/// Shepperd's method: pick the branch whose divisor is largest so the square root never
+/// approaches zero, which is where the naive trace-only formula loses all its precision.
+fn quaternion_from_rotation(r: [[f64; 3]; 3]) -> Quaternion {
```
edgarriba: @gbin for this conversion stuff i suggest to use glam which is a light lib and provides battle tested geometry conversions

gbin:
the other PR has been merged, please rebase/deconflict

components/payloads/cu_ros2_payloads/src/geometry_msgs.rs
edgarriba: Done in de1b53f — `DQuat::from_mat3` replaces the hand-written Shepperd extraction and `DAffine3::from_rotation_translation` replaces the hand-built homogeneous matrix, about 90 lines of math gone.

Two notes on what I kept: the rigidity check stays explicit rather than using `DAffine3::to_scale_rotation_translation`, because that decomposition assumes its input is scale-times-rotation and so cannot report shear — it is now written in glam ops (`R^T R` vs `DMat3::IDENTITY`, plus `determinant()`). And the row-major/column-major boundary is down to one function, `rotation_matrix`: `Transform3D::rotation` returns rows, glam stores columns.

Since we now use glam directly, `cu-spatial-payloads` keeps its default `glam` feature instead of `default-features = false`, and both pin 0.33.

edgarriba:
@gbin rebased onto master — #1365 was squash-merged, so I replayed only this PR's own commit onto it (`git rebase --onto origin/master`). No conflicts, and the diff is now just this change.

Also switched the geometry to glam per the suggestion in the thread above: `DQuat::from_mat3` and `DAffine3::from_rotation_translation` replace the hand-written quaternion math. The rigidity check stays explicit, since `to_scale_rotation_translation` cannot report shear.

Re-verified after both: 36 tests pass on jazzy and humble, clippy clean, fmt clean, `cu-ros2-bridge` builds.

gbin:
@edgarriba can you set claude to not claim copyright over the codebase? thanks

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Gn1oe6gBfhFHenJBbo7nXN

edgarriba:
@gbin done — turned off at the tool level, so it will not come back on future contributions.

This branch is force-pushed with both trailers stripped from both commit messages (same tree, messages only), and I removed the generated-by footer from the PR description. For the record those trailers are git attribution metadata rather than a copyright assertion — Anthropic does not claim ownership of output — but happy to keep them out of the history either way.

One I cannot fix from here: #1365 was already squash-merged, and GitHub carried the trailers into 525cfcb on master. Rewriting that is yours to decide, and probably not worth it.

PR #1371, opened 2026-09-08
feat(cu-ros2-payloads): add JointState, BatteryState and tf2_msgs/TFMessage

edgarriba:
## What this is

Three ROS 2 message types that `cu-ros2-payloads` still lacks, added as wire structs with their Jazzy RIHS01 hashes: `tf2_msgs/TFMessage`, `sensor_msgs/JointState` and `sensor_msgs/BatteryState`.

This comes out of running Copper on an **XLeRobot** — a Jetson Orin, an OAK-D doing stereo VIO, a Feetech STS3215 servo base and two SO-101 arms — publishing to `rmw_zenoh` through `cu-ros2-bridge`. Those are the three topics the robot publishes that this crate has no type for: `tf`, `wheel_state` (and the arm joints), and `battery` (the servo bus voltage).

Follow-up to #1365, which added the wire structs and the `RosMessage` identity constants, and #1366, which bridged `cu_spatial_payloads` and added the test that checks every `TYPE_HASH` against `all_rihs.md`. Same shape, same conventions.

## Why `TFMessage` is the one that matters

`/tf` and `/tf_static` carry a `tf2_msgs/TFMessage`, never a bare `geometry_msgs/TransformStamped`, and `tf2_ros` subscribes as `TFMessage` and nothing else.

Because the type hash is part of the rmw_zenoh key expression, a Copper graph publishing a `TransformStamped` on `/tf` today is not merely publishing the wrong shape — it is publishing on a key no transform listener is subscribed to. Nothing errors at either end. `rviz2` shows an empty tf tree and reports every frame as non-existent, and `tf2_ros` never sees a transform to buffer.

So as things stand, **a Copper graph cannot publish a transform tree that ROS will consume at all**, whatever it puts in the message. That is what this type fixes, and it is why I would still want this PR even if the other two were dropped.

The wrapping is also what makes a tree possible rather than being incidental packaging: tf2 requires that all the transforms making up one update arrive together, which is why the message is a sequence of edges rather than one message per edge.

## The other two

**`sensor_msgs/JointState`** — what any robot with joints publishes. The wire subtlety is that the four arrays are parallel to `name` and each may be empty, which the definition means as *not reported* rather than *zero*: a platform that measures position and velocity but no torque sends an empty `effort`. A test pins that an empty sequence still writes its `uint32` length.

**`sensor_msgs/BatteryState`** — deliberately derives no `Default`. Every float after `voltage` documents NaN as its unmeasured value, so a zeroed message claims a flat battery at 0 °C instead of saying nothing, and the three `power_supply_*` enums have an explicit `_UNKNOWN = 0` so that unknown gets stated rather than defaulted into. `Imu` and `Temperature` already derive no `Default` here, so this follows the module rather than departing from it. The enums are exposed as associated constants because the fields are bare `u8`s — without them every publisher writes a magic number and every consumer compares against one.

## Provenance of the hashes

**All three hashes are taken from this repository's own `components/payloads/cu_ros2_payloads/all_rihs.md`**, not from anywhere else. I did not compute or copy them from another tree. #1366's `type_hashes_match_the_published_rihs_table` test now covers all 23 entries and checks each against that table via `include_str!`.

Field names, order and types were verified against the Jazzy `.msg` definitions in `ros2/common_interfaces` (`sensor_msgs/msg/JointState.msg`, `sensor_msgs/msg/BatteryState.msg`) and `ros2/geometry2` (`tf2_msgs/msg/TFMessage.msg`).

## Tests

The layout tests go past round-tripping on purpose. A round trip through the same struct cannot detect a field order that disagrees with the `.msg`: swap `charge` and `capacity`, or `present` and `power_supply_status`, and the trip is still symmetric, the length unchanged and the type hash unchanged — while every ROS consumer decodes the wrong field with no error. So `BatteryState`'s fixed-size region is asserted byte for byte, pinning that the three enums and `present` occupy four consecutive unpadded bytes (CDR has no packed bool), and `TFMessage`'s sequence length prefix is checked at its offset.

## No adapters

None of the three gets a `RosMsgAdapter`. There is no Copper payload for a joint set, a battery or a transform tree to bind them to, and `TFMessage` and `JointState` would additionally need a deliberate source for the header — the same reason #1366 left the stamped messages unbound. They are directly usable by a bridge today.

## Verification

Run on the Orin, in a fresh clone branched from upstream `master`:

- `cargo test -p cu-ros2-payloads --features jazzy` — **41 passed; 0 failed** (36 before this change, 5 new).
- `cargo test -p cu-ros2-payloads --features humble` — **41 passed; 0 failed**.
- `cargo clippy -p cu-ros2-payloads --features jazzy --all-targets -- -D warnings` — clean.
- `cargo fmt -p cu-ros2-payloads -- --check` — clean.

I could not run `just ros-rihs01-hashes`: it reads `/opt/ros/jazzy/share` and there is no ROS install on this machine, which is exactly why the hashes come from the checked-in table instead.

To be explicit about what is *not* verified here: the robot publishes these three message types today, but through a separate hand-written implementation, not through this crate. The wire layouts agree; these particular structs have not themselves been on that robot's wire.

### DIFF
--- a/components/payloads/cu_ros2_payloads/src/lib.rs
+++ b/components/payloads/cu_ros2_payloads/src/lib.rs
@@ -4,6 +4,7 @@ pub mod geometry_msgs;
 pub mod nav_msgs;
 pub mod sensor_msgs;
 pub mod std_msgs;
+pub mod tf2_msgs;
 
 #[cfg(all(feature = "humble", feature = "jazzy"))]
 compile_error!("features `humble` and `jazzy` are mutually exclusive");
@@ -174,14 +175,17 @@ mod tests {
             entry::<crate::geometry_msgs::Vector3>(),
             entry::<crate::nav_msgs::Odometry>(),
             entry::<crate::nav_msgs::Path>(),
+            entry::<crate::sensor_msgs::BatteryState>(),
             entry::<crate::sensor_msgs::CameraInfo>(),
             entry::<crate::sensor_msgs::CompressedImage>(),
             entry::<crate::sensor_msgs::Image>(),
             entry::<crate::sensor_msgs::Imu>(),
+            entry::<crate::sensor_msgs::JointState>(),
             entry::<crate::sensor_msgs::MagneticField>(),
             entry::<crate::sensor_msgs::PointCloud2>(),
             entry::<crate::sensor_msgs::PointField>(),
             entry::<crate::sensor_msgs::RegionOfInterest>(),
+            entry::<crate::tf2_msgs::TFMessage>(),
         ]
     }
 
--- a/components/payloads/cu_ros2_payloads/src/sensor_msgs.rs
+++ b/components/payloads/cu_ros2_payloads/src/sensor_msgs.rs
@@ -193,6 +193,119 @@ impl RosMessage for MagneticField {
         "RIHS01_e80f32f56a20486c9923008fc1a1db07bbb273cbbf6a5b3bfa00835ee00e4dff";
 }
 
+// sensor_msgs/JointState
+//
+// The arrays are parallel to `name` and each one may be empty, which the message defines as "not
+// reported" rather than "zero": a robot that publishes positions but measures no effort sends an
+// empty `effort`, not a run of zeros. Anything else must have the same length as `name`, since
+// that is the only thing associating a value with a joint.
+//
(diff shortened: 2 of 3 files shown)


gbin:
We will need to bring the internal copper stuff up to speed to support those ROS2 concepts. paging @makeecat ;)

PR #1380, opened 2026-09-10
Frame logging codec payloads directly in Copper storage

gbin:
A codec decoder shares the CopperList stream with other payloads and metadata. Copper now gives every present codec payload its own byte boundary: reserve a fixed four-byte little-endian length, encode directly into the destination, measure the writer cursor, and fill in the reserved header. Decoding caps reads, short reads, peeks, and consumption at that boundary and rejects incomplete consumption.

The fixed-storage path invokes the codec once and adds no payload allocation or intermediate payload copy. Writer wrappers preserve backpatching and count the reserved header once. PNG retains direct encoding; its decoding adapter also supports the non-peekable reader used by log export through a fixed 1 KiB buffer.

This changes encoded codec payload content. Recordings require the matching application logreader; the unified-log encapsulation version remains unchanged. Custom codec writers need the new bincode position/overwrite operations. Both the workspace and standalone RP2350 example pin the tested cu-bincode commit pending its 2.2 release.

Validation:
- `just codec-framing-check`: framing regressions, adjacent PNGs with pixel checks through slice and streaming readers, generated replay, storage/accounting tests, clippy, and shared no_std checks.
- `just nostd-ci`: passes, including 596 tests and embedded/RP2350 firmware and host builds.
- `just api-check`: passes with the two intentional ObservedWriter trait methods reflected in the snapshot.
- Optional FFmpeg-enabled FFV1 tests are blocked in `ffmpeg-next` by API mismatches with the installed FFmpeg headers. Default FFV1 tests pass.

Replaces #1301. Supporting dependency: https://github.com/copper-project/cu-bincode/pull/9 (CI passes). Book documentation: https://github.com/copper-project/copper-rs-book/pull/69 (`just build` passes). The wiki source update is on its `docs/codec-payload-framing` review branch; GitHub wikis do not support pull requests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -292,7 +292,7 @@ cu-ros2-payloads = { path = "components/payloads/cu_ros2_payloads", version = "1
 cu-pid = { path = "components/tasks/cu_pid", version = "1.2.0-dev" }
 
 # External serialization
-bincode = { package = "cu-bincode", version = "2.1", default-features = false, features = [
+bincode = { package = "cu-bincode", version = "2.2", git = "https://github.com/copper-project/cu-bincode", rev = "f10d7ac8b061cf63e9b69a0a13c76919a3a4599a", default-features = false, features = [
   "derive",
   "alloc",
 ] }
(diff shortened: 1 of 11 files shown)


gbin:
This is absolute trash closing

PR #1391, opened 2026-09-11
feat(unifiedlog): add rollover(size) for fixed-size on-disk logging

### LINKED ISSUE #1377: feature: Implement a "rollover" unified log, ie fixed size log on disk with only the last logs.
## Problem statement

See #1376 today we just assume that you have enough space for a full robot run. This might be true for an experimental autonomous robot but usually in prod you start to cut off disk space for cost reason for example.

## Proposed solution

Start to have retention policies in the log starting with the simplest one: just set a maximum size for the log and only keep that maximum size of the last available logs.

This combines well with the append(true) for the logs that are at the time of writing in another PR.

In the API I would just add rollover(size) as an option. We will need to think carefully about the restarts with changed parameters etc.

webzuweb:
Implements #1377 — a retention policy for the unified log, starting with the simplest one: a maximum size on disk, keeping only the last logs.

## What it does

The `MmapUnifiedLoggerBuilder` gains a `rollover(size)` option (requires `write` + `create`). It turns the slab files into a ring of `max(1, size / slab_size)` slabs; once the ring is full, the oldest slab is overwritten so only the most recent logs are retained.

## Design (aligned with @gbin in #1377)

- **`MainHeader` extended + `UNIFIED_LOG_FORMAT_VERSION` bumped to 2.** The header now carries:
  - `head_slab_index` — first retained slab (0 for `no_std`),
  - `head_offset` — byte offset of the first retained section (within-slab for `std`; absolute bytes for `no_std`),
  - `max_slabs` — ring size (0 = unbounded),
  - `slab_size` — bytes per slab (so `append(true)` can detect geometry mismatch).
- **Reader starts at the head pointer** and wraps around the ring (`next_slab` modulo `max_slabs`), stopping at the end-of-log marker. A safety guard errors out if a full ring is traversed without finding the marker.
- **`append(true)` fails loudly** when the requested ring geometry (slab size or rollover size) does not match the existing log.
- The `no_std` backend (`cu_sdlogger`) writes the v2 header with the head pointer in absolute bytes (`head_offset = BLK`), no ring yet (`max_slabs = 0`).

## Tests

- `test_rollover_keeps_only_last_slabs` — writes enough to wrap a 2-slab ring, asserts only 2 slab files exist and that readback returns a non-empty, strictly smaller history.
- `test_append_rollover_geometry_mismatch_fails` — asserts `append(true)` errors with `InvalidInput` on mismatched slab size.

`cargo test -p cu29-unifiedlog` → 16 passed; `cargo clippy -p cu29-unifiedlog --all-features` clean.

Closes #1377.

### DIFF
--- a/core/cu29_unifiedlog/src/memmap.rs
+++ b/core/cu29_unifiedlog/src/memmap.rs
@@ -101,6 +101,7 @@ pub struct MmapUnifiedLoggerBuilder {
     write: bool,
     create: bool,
     append: bool,
+    rollover_size: Option<usize>,
 }
 
 impl Default for MmapUnifiedLoggerBuilder {
@@ -117,6 +118,7 @@ impl MmapUnifiedLoggerBuilder {
             write: false,
             create: false, // This is the safest default
             append: false,
+            rollover_size: None,
         }
     }
 
@@ -151,6 +153,15 @@ impl MmapUnifiedLoggerBuilder {
         self
     }
 
+    /// When `write` and `create` are both set, keep the log bounded on disk:
+    /// slabs form a ring of `max(1, size / slab_size)` files and, once the ring
+    /// is full, the oldest slab is overwritten so only the most recent logs are
+    /// retained.
+    pub fn rollover(mut self, size: usize) -> Self {
+        self.rollover_size = Some(size);
+        self
+    }
+
     pub fn build(self) -> io::Result<MmapUnifiedLogger> {
         let page_size = page_size::get();
 
@@ -167,10 +178,14 @@ impl MmapUnifiedLoggerBuilder {
                     "Preallocated size is required for write mode",
                 )
             })?;
+            let max_slabs = match self.rollover_size {
+                Some(size) => (size / preallocated_size).max(1),
+                None => 0,
+            };
             let ulw = if self.append {
-                MmapUnifiedLoggerWrite::append(&file_path, preallocated_size)?
+                MmapUnifiedLoggerWrite::append(&file_path, preallocated_size, max_slabs)?
             } else {
-                MmapUnifiedLoggerWrite::new(&file_path, preallocated_size, page_size)?
+                MmapUnifiedLoggerWrite::new(&file_path, preallocated_size, page_size, max_slabs)?
             };
             Ok(MmapUnifiedLogger::Write(ulw))
         } else {
@@ -501,6 +516,12 @@ pub struct MmapUnifiedLoggerWrite {
     slab_size: usize,
     /// current suffix for the backing files.
     front_slab_suffix: usize,
+    /// number of slabs in the rollover ring; 0 means unbounded (no rollover).
+    max_slabs: usize,
+    /// index of the oldest retained slab (the first retained section).
+    head_slab_index: usize,
+    /// byte offset of the first retained section within its slab.
+    head_offset: u64,
 }
 
 fn build_slab_path(base_file_path: &Path, slab_index: usize) -> io::Result<PathBuf> {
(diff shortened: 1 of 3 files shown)


gbin:
head_slab_index — first retained slab (0 for no_std),
head_offset — byte offset of the first retained section (within-slab for
std; absolute bytes for no_std),
max_slabs — ring size (0 = unbounded),
slab_size — bytes per slab (so append(true) can detect geometry mismatch).

^^

Here I don't think we need to leak slabs into the copper log format. We
want to be able to have rollover with the partition version of it too (no
std)



head_slab_index can be computed from head_offset and  slab size already
created on disk

slab_size : no need to have it here if there is only one slab on disk, we
just use the slab size set at open, if there are 2+ slabs we assume that
the previous slab size is the one of slab 0 and we bail if we try to open
them with a different parameter.

max_slabs should not appear, the last slab will be of whatever slab size we
have currently have so we can honor rollover_size


So if I am not mistaken we only need head_offset in there.

webzuweb:
Thanks — that's a fair point. Storing the slab-ring geometry in the log format leaks an implementation detail that the no_std/partition backend would need a different shape for. Agreed, let's slim the persisted header to just `head_offset`.

Here's the concrete plan:

- **Persist only `head_offset`** (absolute byte offset of the first retained section over the whole log). Drop `head_slab_index`, `max_slabs`, `slab_size` from `MainHeader`.
- **`head_slab_index`** is derived: `head_offset / slab_size`.
- **`slab_size`** is derived from on-disk geometry: for a single slab, use the size requested at open; for 2+ slabs, take slab 0's size and bail on mismatch (exactly as you described).
- **`max_slabs`** is derived at open: `rollover_size / slab_size`, never persisted.

One thing I want to confirm before I rework the reader: without a persisted `max_slabs`, the read side needs to know where the ring ends to (a) wrap correctly when replaying across a wrapped ring and (b) fail cleanly instead of looping. My plan is to have the reader derive the ring extent from the set of slab files actually present on disk (they're created 0..N in order and then reused), using the end-of-log marker as the authoritative stop. Does that match your mental model, or would you prefer the reader to take `rollover_size` (or the slab count) as an explicit open parameter instead?

I'll push the refactor with tests once I've confirmed this point.

gbin:
Yes let's do this thanks

webzuweb:
Done — slimmed the persisted header as agreed.

**MainHeader** now carries only `head_offset` (absolute byte offset of the first retained section). Dropped `head_slab_index`, `max_slabs`, and `slab_size` from the on-disk format.

- **Writer**: keeps its in-memory ring fields and serializes the absolute head offset via `absolute_head_offset()` = `head_slab_index * slab_size + head_offset`.
- **Rollover rings no longer trim slab files on drop**, so every slab stays at its full preallocated size and `slab_size` is derivable from slab 0's on-disk size (linear logs keep trimming).
- **Reader**: derives `(slab_size, ring_extent)` from the contiguous slab files on disk (`discover_slab_geometry`) and splits the absolute head pointer (`head_slab = head_offset / slab_size`), using the end-of-log marker as the authoritative stop.
- **append(true)**: validates slab size only for rings (where it's derivable); a linear log re-applies the requested preallocated size on resume.

Verified: `cargo test -p cu29-unifiedlog --lib` 16 passed, clippy/fmt clean, `cargo check -p cu-sdlogger` ok (no_std backend updated to the slimmer header).

gbin:
@webzuweb updated the API, looks ok

core/cu29_unifiedlog/src/memmap.rs
```diff
@@ -1273,6 +1273,13 @@ impl MmapUnifiedLoggerRead {
             self.current_slab_index = pos.slab_index;
         }
         self.current_reading_position = pos.offset;
+        // Count the destination's distance from the head, so rereading a slab
```
gbin, line 1276: @webzuweb CI/CD brought up this bug, are you ok with the fix?

webzuweb:
@gbin yes, that fix is correct — good catch from CI/CD, I'm fine with it as-is.

The root cause: `slabs_read` was previously only ever incremented inside `raw_skip_section`, so after a `seek()` back to a saved position the counter stayed stale. Re-reading a slab from that position kept accumulating against `max_slabs`, so on a ring the `slabs_read >= max_slabs` guard could trip early and end the scan prematurely even though the ring hadn't actually been exhausted. Re-deriving it from the destination's distance to `head_slab_index` — including the wraparound branch for `pos.slab_index < head_slab_index` — matches the ring geometry exactly.

The new `test_seek_and_reread_across_slab_boundaries` covers the precise seek → re-read path that was broken (it does the `raw_skip_section` to discover, then `seek` back and `raw_read_section`, asserting the count is stable across two passes). Thanks for pushing the fix.

PR #1399, opened 2026-09-17
fix(cu-ros2-bridge): advertise best-effort QoS instead of an empty profile

### LINKED ISSUE #1398: cu_ros2_bridge advertises an empty QoS string, so subscribers infer defaults the bridge does not honour
`Topic::qos()` in `components/bridges/cu_ros2_bridge/src/topic.rs` returns a hardcoded empty QoS
string for every channel:

```rust
pub fn qos(&self) -> &str {
    // TODO implement QoS
    "::,:,:,:,,"
}
```

That string goes into the rmw_zenoh liveliness token (`liveliness.rs:37,49`), which is where a ROS 2
subscriber reads the publisher's QoS during discovery.

## Why it is more than a missing feature

rmw_zenoh encodes QoS in that token as
`reliability:durability,history:depth,deadline:lifespan,liveliness:lease`. With every field blank, a
subscriber falls back to defaults — and the ROS 2 default profile is `RELIABLE` + `KEEP_LAST(10)`.

The bridge's actual behaviour is best-effort. `send` (`lib.rs:617`) calls
`zenoh::Wait::wait(publisher.put(..))` without setting congestion control, and zenoh's default is
`CongestionControl::Drop` — documented as *"When transmitting a message in a node with a full queue,
the node may drop the message."* So a full queue discards silently.

So the bridge can advertise reliability it does not provide. A subscriber that reasonably expects
retransmission gets silent drops instead, and ROS's two mechanisms for noticing — QoS-incompatibility
at connect time, and `sample_lost` events — are both defeated, because the publisher never declared
anything to be incompatible with.

The drop itself is the right default for sensor streams; ROS makes the same trade in
`rmw_qos_profile_sensor_data` (`BEST_EFFORT` + `KEEP_LAST(5)`). The problem is only that the
contract and the behaviour disagree.

## Where it shows up

We publish a 128 kB depth image at 15 Hz over this bridge alongside camera info, odometry, tf and
compressed video. On a healthy LAN nothing drops. Over a slower link frames start disappearing with
no local counter and no subscriber-side event — and a consumer reading the advertised QoS has been
told that cannot happen.

## Suggested fix

Let a channel declare its QoS and encode it into the token, defaulting sensor-ish streams to
`BEST_EFFORT` + `KEEP_LAST(n)` so the advertisement matches `CongestionControl::Drop`. Even
hardcoding best-effort for all Tx channels would be more accurate than the current blank, though
per-channel is what ROS consumers will eventually expect.

Happy to send a PR if you can say which shape you would take — a per-channel QoS field in the RON,
or a single accurate default to start with.

webzuweb:
Fixes #1398.

## Problem

`Topic::qos()` returned a hardcoded empty QoS chunk `"::,:,:,:,,"` — the rmw_zenoh encoding for "all defaults". A ROS 2 subscriber reads the publisher's QoS from the liveliness token during discovery; with every slot empty it falls back to the ROS default profile (`RELIABLE` + `KEEP_LAST(10)`).

But the bridge actually publishes **best-effort**: `send` (`lib.rs:617`) calls `zenoh::Wait::wait(publisher.put(..))` without setting congestion control, and zenoh's default is `CongestionControl::Drop` ("the node may drop the message when the queue is full"). So the advertisement claims retransmission the bridge cannot provide, which defeats both QoS-incompatibility detection at connect time and `sample_lost` events — frames silently vanish on a slow link while the subscriber was told that can't happen.

## Fix

Encode an accurate profile into the token, matching the bridge's actual behaviour. The new `qos` module mirrors `rmw_zenoh`'s `qos_to_keyexpr()` bit-for-bit (verified against `rmw_zenoh_cpp/src/detail/liveliness_utils.cpp`), and `Topic` now carries a typed `Qos` that defaults to a sensor-data profile:

* `BEST_EFFORT` + `KEEP_LAST(5)` (mirroring `rmw_qos_profile_sensor_data`) → encodes to `"2::,5:,:,:,,"`

Deadline, lifespan, liveliness and lease stay at the rmw_zenoh defaults (encoded as empty slots), which is accurate for the bridge today.

I introduced a typed `Qos` rather than another magic string so that per-channel QoS (the eventual goal of #1398) can build on it.

## Tests

* `qos::tests::sensor_data_profile_matches_rmw_zenoh_encoding` — best-effort profile encodes to `"2::,5:,:,:,,"`.
* `qos::tests::all_rmw_defaults_encode_to_the_empty_profile` — full-default profile encodes to `"::,:,:,:,,"` (i.e. the old hardcoded value), proving the encoder matches rmw_zenoh's reference.

Local checks: `cargo test -p cu-ros2-bridge` (7 lib + 1 ros2_interop passed), `cargo clippy -p cu-ros2-bridge --all-targets -- -D warnings` clean, `cargo fmt --check` clean.


gbin:
@edgarriba do you mind giving it a try? LGTM but I want to be sure it works with your real ROS system. Thanks

components/bridges/cu_ros2_bridge/src/qos.rs
```diff
@@ -0,0 +1,145 @@
...
+//! `qos_to_keyexpr()` from `rmw_zenoh_cpp/src/detail/liveliness_utils.cpp`
+//! bit-for-bit, so that a ROS 2 subscriber reads the publisher's real QoS
+//! during discovery instead of falling back to defaults the bridge does not
+//! honour.
+//!
+//! Layout (each value is the numeric RMW QoS policy enum; a slot is left empty
+//! when it equals the rmw_zenoh default):
+//!
+//! ```text
+//! reliability:durability,history:depth,deadline_sec:deadline_nsec,lifespan_sec:lifespan_nsec,liveliness:lease_sec:lease_nsec
```
edgarriba, line 14: is it really needed this text ?

edgarriba (commented):
make sure is somehow tested in a real device

gbin:
> make sure is somehow tested in a real device

but does the fix work for your bug at least? Thanks

webzuweb:
Yes — the fix resolves the reported bug, which was purely about the contract/behaviour mismatch, not the drop behaviour itself.

Before, `Topic::qos()` returned the blank `"::,:,:,:,,"` token, which rmw_zenoh decodes as "every field default" → a subscriber infers `RELIABLE + KEEP_LAST(10)`, even though the bridge actually publishes best-effort (`CongestionControl::Drop`). That silently defeated both of ROS's safety nets: QoS-incompatibility detection at connect time, and `sample_lost` events — because the publisher never declared anything to be incompatible with.

The PR makes the advertisement honest: `Topic` now carries `Qos::sensor_data()` = `BEST_EFFORT + KEEP_LAST(5)`, encoded to the token `"2::,5:,:,:,,"` — the same shape `rmw_qos_profile_sensor_data` produces. So a subscriber now sees best-effort and can (correctly) refuse to connect or surface `sample_lost` if it demanded reliability.

What I *can't* claim from here is real-device verification — I don't have a live ROS 2 subscriber + `rmw_zenoh` stack to watch discovery end-to-end. That's exactly the gap @edgarriba's test fills. Unit tests pin the token *encoding* bit-for-bit against `rmw_zenoh_cpp`'s `qos_to_keyexpr()` (verified against `liveliness_utils.cpp` source): `sensor_data_profile_matches_rmw_zenoh_encoding` and `all_rmw_defaults_encode_to_the_empty_profile`. So the advertised token is provably right; the discovery-time behaviour on real hardware is the part I can only hand off.

gbin:
Can we have at least either @edgarriba that pushed the bug report confirm that it is fixing his issue or @webzuweb testing if this actually works with an actual instance of ROS2?

gbin:
ok no response but it sounds ok, I guess we will merge that and hear from the ROS users if it broke the world. :shrug:

PR #1407, opened 2026-09-18
feat(zenoh-bridge): per-channel queue_mode, so an Rx channel can keep the newest sample

edgarriba:
# PR C — `cu_zenoh_bridge`: per-channel `queue_mode`, so an Rx channel can keep the NEWEST sample

Branch: `feat/zenoh-bridge-queue-mode`, based on `upstream/master` (`aeda9d77c`).
Patch: `C-zenoh-bridge-queue-mode.patch`.
Touches: `components/bridges/cu_zenoh_bridge/src/{lib.rs}`, `README.md`.

**Unlike PRs A and B in this directory, this one is COMPILED AND TESTED**: `cargo check -p
cu-zenoh-bridge --all-targets` and `cargo test -p cu-zenoh-bridge` both pass on the branch.

## The defect

`ZenohBridge::receive` consumes **at most one sample per call**, i.e. one per graph iteration
(`lib.rs`, `try_recv`). Every subscriber is declared with no handler override:

```rust
let subscriber = zenoh::Wait::wait(session.declare_subscriber(key_expr))
```

which is zenoh's `DefaultHandler` — `FifoChannel::default()`, depth
`API_DATA_RECEPTION_CHANNEL_SIZE` (256).

So a channel whose publisher is faster than the consuming graph's rate falls behind, and it falls
behind **losslessly and without bound**: the consumer keeps reading ever-older samples, in order,
with nothing dropped and no error on either side.

## Measured

Two processes on one host, one explicit TCP endpoint, a bare `declare_subscriber` reproducing the
bridge's own declaration. 724 B payloads at 200 Hz, sequence number in the first 8 bytes.

| subscriber | publisher achieved | consumer received |
|---|---|---|
| none | 197.2 Hz | — |
| drains 2000/s (control) | 196.9 Hz | 2952 of 2954 |
| drains 0/s | 197.0 Hz | 0 |
| **drains 100/s** | **196.9 Hz** | **seq 0..2490, contiguous, 0 dropped** |

At 100/s against 200 Hz the consumer was at sequence 2490 while the publisher had reached 5908 —
**12 s behind and growing linearly**. Lag grew 4.9 s (t=10 s) to 12 s (t=25 s).

Two things worth separating, because the obvious reading of zenoh's own doc comment gets this
backwards:

1. **The publisher is never slowed.** Full rate in every case, including against a consumer
   draining nothing for 25 s. `FifoChannel`'s doc warns that a full queue "will block until a slot
   is available … a slow subscriber could block the underlying Zenoh thread" — it does, but that
   thread is the *subscriber's*, and the stall does not cross the process boundary. This is a
   staleness problem, not a back-pressure one.
2. **It is invisible to the consumer.** The channel is never empty, so an "am I receiving
   samples?" check passes; nothing is dropped, so a sequence-gap check passes too. The only symptom
   is that the data is old, which nothing in the bridge surfaces.

The 256-deep FIFO is not the whole story: the consumer received ~2200 more samples than fit in it,
so the rest was buffered upstream in zenoh's transport queues and the socket. That is also why
`put` never waits, and it means a wedged consumer costs the **producer** memory too.

## The change

Per-channel `queue_mode`, mirroring what `cu_ros2_bridge` already exposes
(`cu_ros2_bridge/src/lib.rs`, `RxQueueConfig`) so the two bridges are configured the same way:

- `queue_mode: "fifo"` (default) — unchanged behaviour, ordered and lossless.
- `queue_mode: "ring"` — `RingChannel`, drops the oldest when full, so `receive` yields the newest
  sample available. `ring_size` sets depth, default `1` (latest-wins).

`ZenohSubscriber` becomes a two-variant enum because `FifoChannelHandler` and `RingChannelHandler`
are distinct types and the choice cannot live behind a `dyn`; `start`, `receive` and `stop`
dispatch over it. This is the same shape `Ros2Subscriber` already has.

### Two deliberate differences from the `cu_ros2_bridge` precedent

Both are small and both close a silent-failure hole rather than add surface:

- **`queue_mode` on a Tx channel is refused at construction**, not ignored. On the sending side it
  reads like it bounds the publisher and does nothing whatsoever — the kind of config that looks
  applied for months.
- **`ring_size: 0` is refused at config time.** `RingChannel::new(0)` fails inside zenoh at declare
  time, which is a bewildering place for a typo in a RON to surface.

## Tests

`cu_zenoh_bridge` had no test module at all; this adds one covering `RxQueueConfig::from_config`:
absent config, config without the key, explicit `fifo`, `ring` defaulting to depth 1, explicit
`ring_size`, `ring_size: 0` refused, and an unknown value refused rather than silently falling back
to `fifo` — that last one matters, because a misspelled `queue_mode` quietly restoring the default
is exactly the failure this option exists to prevent, arriving through the config instead.

`cu_ros2_bridge` has no tests for its equivalent, so these are new coverage for the pattern in both
bridges.

## Compatibility

None of this changes default behaviour: a channel with no `queue_mode` gets the same
`DefaultHandler` it gets today. The only new error paths are on configs that are currently
misspelled or meaningless.

## Reproducing the measurement

The harness is `crates/cufx-zenoh/examples/fifo_probe.rs` in `onirik-labs/xlerobot-rs`. It is a
standalone raw-zenoh probe rather than a Copper app, so it reproduces the bridge's declaration
without the runtime:

```
fifo_probe sub --drain-hz 100 --endpoint tcp/127.0.0.1:7470
fifo_probe pub --rate-hz 200 --endpoint tcp/127.0.0.1:7470
```

It can be contributed here as a benchmark if that would be useful; `benchmarks/
cu_zenoh_bridge_bench` is the natural home, and it already measures latency on the same shape.


edgarriba:
@gbin if there's any other way to do this, let me know and will close this

gbin:
It looks correct and copper should preserve the determinism in distributed resim behind that. 

1. maybe add a minimal determinism test to be sure that we can resim when the ring starts to leak
2. If 1 works then default to ring instead of fifo as it is safer in general

Thanks!

edgarriba:
Both done. Short version: a leaking ring cannot break distributed resim, and the reason is
structural rather than lucky — so I've defaulted to `ring`.

## 1. Determinism when the ring leaks

Two independent mechanisms make the leak invisible to resim:

**No subscriber runs at replay time.** `gen_recorded_replay_support` emits an `ExecutedBySim` arm
for every node — tasks, `BridgeTx` and `BridgeRx` alike — each copying its value straight out of
the recorded CopperList (`cu29_derive/src/lib.rs:1694`, and the generated doc says it: *"every
recorded task and bridge output is copied from the CopperList and the runtime implementation is
skipped"*). `queue_mode` therefore has no effect during resim at all; it only shapes what gets
recorded.

**A drop removes a provenance edge, it never dangles one.** `copperlist_origins` is a `filter_map`
over `msg.metadata().origin()`, and on an empty tick `receive` does `clear_payload()` +
`clear_origin()`. So a leaked sample contributes nothing, and the causal graph built in
`plan_nodes` is a strict subgraph of the fifo one. A subgraph of a DAG is still a DAG.

Worth noting the failure you were probing for is the *opposite* shape and already has a test:
`engine_reports_unresolved_recorded_provenance` catches a consumer that recorded an origin whose
producer log is missing. Dropping on the wire can't produce that — the consumer never saw the
message, so it never wrote the edge.

**New test:** `engine_replays_deterministically_when_a_ring_rx_leaks` in
`cu29_runtime::distributed_replay`. The fixture models both halves of a leak —
`RING_DRAINS = [Some(0), None, Some(3), None, None, Some(5)]`, i.e. samples overwritten between
drains *and* ticks where the ring was empty — then asserts every recorded CopperList still
replays (a producer node nothing consumed is a leaf, not an orphan that stalls the frontier),
that surviving edges still order producer-before-consumer, and that a second independently built
plan replays in the identical order.

The regression it actually pins: if validation were ever tightened to require every producer
CopperList to be consumed, ring mode would deadlock the frontier. Today it doesn't.

**Also ran it end to end** on `cu_distributed_resim_demo`, with the scout ticking at 500 Hz
against the 100 Hz planners so the pulse ring genuinely overruns:

```
scout    --period-ms 2  --iterations 270   (x2 instances)
plan     --period-ms 10 --iterations 120   (x2)
control  --period-ms 10 --iterations 120   (x2)

distributed replay verified 6 subsystem logs across 1020 causal nodes
```

Byte-identical CopperList and keyframe streams on all six logs. To be precise about what is
measured and what is derived: the resim verification is measured; the drop volume is arithmetic —
each scout emits on every third CopperList, so 180 pulses cross the shared route in 540 ms while
each planner drains at most one per 10 ms tick (~54), leaving ~70% dropped. I kept the consumers
alive past the end of the scout window, otherwise the "drops" would just be a dead subscriber.

## 2. Ring is now the default

`RxQueueConfig::DEFAULT = Ring { size: 1 }`. Rationale in the doc comment: unbounded staleness is
silent and a drop is not, which is the asymmetry that makes ring the safer default.

One adjacent bug fixed while there: `ring_size` was only read under an explicit `queue_mode`, so a
channel configured with `ring_size: 8` alone got a fifo and no diagnostic. It now sets the depth on
its own. `fifo` remains reachable by spelling it out, and an unknown `queue_mode` is still an error
rather than a fallback.

## One question back

`cu_ros2_bridge` already ships this same `queue_mode`/`ring_size` option (`lib.rs:140`), also
defaulting to fifo. Flipping only the zenoh bridge leaves the two disagreeing on the same key. I've
deliberately not touched it — changing a second component's public default inside this PR felt like
scope you hadn't asked for — but happy to do it here or in a follow-up, whichever you prefer. Its
version also lacks the `ring_size == 0` guard, which panics inside `RingChannel::new` at declare
time.

PR #1408, opened 2026-09-18
feat(ros2-payloads): nav_msgs/OccupancyGrid and nav_msgs/MapMetaData

edgarriba:
Adds `nav_msgs/OccupancyGrid` and `nav_msgs/MapMetaData` to `cu_ros2_payloads`.

## Why

`nav_msgs` currently has `Odometry` and `Path`. These two are the remaining `nav_msgs` types a mobile robot needs — what a costmap, a SLAM node and RViz all speak — so any downstream publishing a map has to define them itself. I hit exactly that building an occupancy grid on a Jetson Orin and would rather not maintain a private copy of a standard message.

Field layout is the jazzy definition:

```
MapMetaData   = (map_load_time, resolution, width, height, origin)
OccupancyGrid = (header, info, data)
```

`MapMetaData` derives `Copy` — five scalars and a `Pose`.

## Two traps, both pinned by tests

Both fail by **decoding cleanly into the wrong thing**, which is why they are worth the tests rather than a comment:

**`data` is `Vec<i8>` and deliberately NOT `serde_bytes`.** `serde_bytes` takes `u8` only, so an `int8[]` encodes element by element rather than as the single octet sequence `CompressedImage` gets. The types differ in signedness, not width — so a `u8` reinterpretation round-trips perfectly and turns every unknown cell (`-1`) into `255`. A consumer clamping to `0..=100` then reads never-observed space as *certainly occupied*; clamped the other way, it reads it as free space nobody has looked at. `the_unknown_cell_sentinel_survives_the_wire` pins it.

**`data.len()` must equal `width * height`.** Nothing in the type enforces it, and a row-major consumer either walks off the end or renders a torn map.

## Hashes

The RIHS01 literals are the jazzy ones, consistent with the `Odometry` and `Path` literals already in this file.

The `OccupancyGrid` hash is verified **on the wire**, not only against a table — it is live on a robot here, published and consumed through rmw_zenoh as:

```
0/<robot>/map/nav_msgs::msg::dds_::OccupancyGrid_/RIHS01_8d348150c12913a31ee0ec170fbf25089e4745d17035792a1ba94d6f0bc0cfc7
```

`MapMetaData` is nested inside `OccupancyGrid` and never published alone, so only the outer hash is exercised end to end.

## Tests

`cargo test -p cu-ros2-payloads` passes, including four new cases: a full round-trip with every class of cell (unknown / free / partial / full), the sentinel test above, the dimension invariant, and the existing suite unchanged.


gbin:
Note: please run "just fmt" on your PRs thanks

PR #1409, opened 2026-09-18
feat(planner): add portable exact schedules + no lock lane sync

gbin:
## Summary

This is the new generalized planner for copper accepting distributed copperlists and lockless coordination inter CL.

## Related issues
- Closes #

## Reminder


gbin:
Ok this is not completely where I want it to be but this is a safe intermediate step:

gbin:
what is missing: pipeline should optimize adjacent stateless tasks + one option to NOT measure timing in between 2 fused tasks

